PR #22, opened 2022-07-18
replace local references

cgenesoniSouthWorks:
replacing some local references on package folder.
Added StringContact file.


constants/StringConstants.js
```diff
@@ -0,0 +1,33 @@
+const StringConstants = {
```
seanspeaks, line 1: @cgenesoniSouthWorks I think we can just delete the reference to this and remove the file altogether. Which file is requiring this?

seanspeaks (commented):
Feels like we can merge but want to confirm on the StringConstants.js file first

seanspeaks (approved):

PR #97, opened 2023-01-19
wip: trying to throw a helpful error

joncodo:
If you don't have AES_KEY and AES_KEY_ID set, the encrypt method fails.

It took me quite a long time to figure this out as a new developer so I think a helpful error message here goes a long way.

I can't seem to get my test to pass even when I pass in the keys in the command line like this:

`AES_KEY=Testing123456789Testing123456789 AES_KEY_ID=xxx-xxx-xxx-xxx-xxx jest /Users/jon/Code/lefthook/frigg/packages/encrypt/Cryptor.test.js`

I need some help to get this merged 😄

### DIFF
--- a/packages/encrypt/Cryptor.js
+++ b/packages/encrypt/Cryptor.js
@@ -20,7 +20,7 @@ class Cryptor {
     }
 
     async generateDataKey() {
-        if (this.shouldUseAws) {
+        if (this?.shouldUseAws) {
             const kmsClient = new AWS.KMS();
             const dataKey = await kmsClient
                 .generateDataKey({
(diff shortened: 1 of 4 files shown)


joncodo:
Honestly, maybe AES configuration should not be a part of frigg? I think there can be other ways to encrypt a key without adding the AES dependency?

packages/encrypt/Cryptor.js
```diff
@@ -36,8 +36,15 @@ class Cryptor {
         }
 
         const { AES_KEY, AES_KEY_ID } = process.env;
+        console.log('JON >>> process.env.AES_KEY', process.env.AES_KEY);
```
seanspeaks, line 39: Remove the log

packages/encrypt/Cryptor.js
```diff
@@ -36,8 +36,15 @@ class Cryptor {
...
         const randomKey = crypto.randomBytes(32).toString('hex').slice(0, 32);
 
+        if (!AES_KEY || !AES_KEY_ID) {
```
seanspeaks, line 42: I _think_ this needs to be `if (!AES_KEY && !AES_KEY_ID)` as we're trying to say "use the key and key ID for manual key rotation"

seanspeaks (commented):
The counter-argument for not requiring AES key for Frigg; we're not. It's optional inclusion if you want field-level encryption on DB read/write, with the ability to use KMS alternatively.

That said, being able to support multiple encryption methods makes sense, though I wouldn't go too overboard. We did AES 256 because it was secure and allowed us to do a manual key rotation

PR #122, opened 2023-02-03
Headers, man. Headers. We should work out where to put the default ty…

seanspeaks:
…pe. Not liking a "per integration, if you're clever enough to put it in the right place"... definitely not liking "per request/method". Anywho.


seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-yotpo@0.0.15`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-yotpo%400.0.15) :rocket:

PR #131, opened 2023-02-15
Switch to Upsert for Credential and Entity creation

seanspeaks:
We should consider doing a fullsale switch to upsert. Not sure if there's a use case where we'd want to find OR create within the processAuthCallback flow... it would always be find and update or create, aka upsert

—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)

### DIFF
--- a/api-module-library/ironclad/manager.js
+++ b/api-module-library/ironclad/manager.js
@@ -72,12 +72,12 @@ class Manager extends ModuleManager {
             debug('No permission to get connection information');
         }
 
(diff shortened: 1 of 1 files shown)


seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-ironclad@0.0.31`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-ironclad%400.0.31) :rocket:

PR #135, opened 2023-03-06
add token_revoked event for slack

debbie-yu:
It looks like we get a `token_revoked` event from slack when we relink Slack (current hypothesis). We see the error when trying to call getUserProfileById.
Adding `token_revoked` event to see if we're able to refresh the token after and get a successful call.

### DIFF
--- a/api-module-library/slack/api.js
+++ b/api-module-library/slack/api.js
@@ -101,7 +101,8 @@ class Api extends OAuth2Requester {
         } else if (
             parsedResponse.error === 'invalid_auth' ||
             parsedResponse.error === 'auth_expired' ||
-            parsedResponse.error === 'token_expired'
+            parsedResponse.error === 'token_expired' ||
+            parsedResponse.error === 'token_revoked'
         ) {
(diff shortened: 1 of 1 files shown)


api-module-library/slack/api.js
```diff
@@ -101,7 +101,8 @@ class Api extends OAuth2Requester {
         } else if (
             parsedResponse.error === 'invalid_auth' ||
             parsedResponse.error === 'auth_expired' ||
-            parsedResponse.error === 'token_expired'
+            parsedResponse.error === 'token_expired' ||
+            parsedResponse.error === 'token_revoked'
```
seanspeaks, line 105: Only thought... `token_revoked`, would definitely not be refreshable. But, doesn't hurt to retry and fail

seanspeaks (approved):

seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-slack@0.1.29`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-slack%400.1.29) :rocket:

PR #149, opened 2023-04-18
add get by id method for pages and blogs

MichaelRyanWebber:
add tests for above and cleanup
make url formation code consistent for pages and blogs

### DIFF
--- a/api-module-library/hubspot/api.js
+++ b/api-module-library/hubspot/api.js
@@ -658,9 +658,18 @@ class Api extends OAuth2Requester {
     // **************************   Pages   *****************************
 
     async getLandingPages(query=''){
-
         const options = {
-            url: `${this.baseUrl}${this.URLs.landingPages}${query}`,
+            url: `${this.baseUrl}${this.URLs.landingPages}`,
+        };
+        if (query !== '') {
+            options.url = `${options.url}?${query}`
+        }
+        return this._get(options);
+    }
+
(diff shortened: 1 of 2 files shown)


api-module-library/hubspot/api.js
```diff
@@ -658,17 +658,26 @@ class Api extends OAuth2Requester {
     // **************************   Pages   *****************************
 
     async getLandingPages(query=''){
+        const options = {
+            url: `${this.baseUrl}${this.URLs.landingPages}`,
+        };
+        if (query !== '') {
+            options.url = `${options.url}?${query}`
```
seanspeaks, line 665: @MichaelRyanWebber 

If you pass a `query` object into the `options`, [we have a function that breaks it down into a query param](https://github.com/friggframework/frigg/blob/1399d648bbb4719750fb9f499824a761adc0d5c9/packages/module-plugin/requester/requester.js#L35)

This may not be ideal, depending on how a query is formed, so 👍 to ignore or not use it.

seanspeaks (approved):
One optional comment, otherwise LGTM 👍 Thanks for the granular requests!

seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-hubspot@0.8.27`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-hubspot%400.8.27) :rocket:

PR #179, opened 2023-06-13
Feature/lef 262 get file details

roboli:


roboli:
SonarCloud scanned test files for duplicate code. I think it is common to have duplicate (repetitive) code on tests? @leofmds @seanspeaks

leofmds:
> @roboli , I believe there is a lot of room for improvement. In a lot of tests we are just doing the same thing, for example, setting up the integration, getting x and y, etc. So this could be put in a function just for that, and you improve the duplicates.
What would need some discussion is if we're doing the same assertions for more than one test. Should we leave it there for clarity, or just call another function that does the assertions?

leofmds (approved):

seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-frontify@0.0.4`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-frontify%400.0.4) :rocket:

PR #182, opened 2023-06-20
hubspot publishing endpoints for pages and blog posts

MichaelRyanWebber:


api-module-library/hubspot/tests/api.test.js
```diff
@@ -299,36 +299,50 @@ describe(`${config.label} API tests`, () => {
...
+            const response = await api.publishLandingPage(
+                sampleLandingPage.id,
+                now.toISOString(),
+            );
+            expect(response).toBeDefined();
+        });
+        it('push a Landing page draft to live' , async () => {
+
+            const response = await api.pushLandingPageDraftToLive(sampleLandingPage.id);
+            expect(response).toBeDefined();
```
hudsontavares, line 344: Is it possible to check a few more details (e.g., HTTP response code, body details expected via `expect(response).toStrictEqual(expect.objectContaining(...))`?

Giving more clues about the expected response here would help to make the target excerpt more resilient to future refactoring.

api-module-library/hubspot/tests/api.test.js
```diff
@@ -299,36 +299,50 @@ describe(`${config.label} API tests`, () => {
...
                 true );
             expect(response).toBeDefined();
         });
+        it('publish a Landing Page' , async () => {
+            const now = new Date(Date.now()+5000);
+            const response = await api.publishLandingPage(
+                sampleLandingPage.id,
+                now.toISOString(),
+            );
+            expect(response).toBeDefined();
```
hudsontavares, line 339: Same question about expected output checking.

api-module-library/hubspot/api.js
hudsontavares, line 1: Minor: given the number of times where we need to repeat `this.baseUrl` in this file, I wonder if we couldn't include something like:

```
// On the base class?
_withBaseUrl(suffix) {
  return `${this.baseUrl}${suffix}`;
}

...


// Invoking places example
async pushLandingPageDraftToLive(objId){
        const options = {
            url: this._withBaseUrl(`${this.URLs.landingPageById(objId)}/draft/push-live`),
        };
        return this._post(options);
    }
```

hudsontavares (commented):
@MichaelRyanWebber - a few comments, please answer them accordingly to the complexity or with a reasoning if you oppose implementing any of them.

api-module-library/hubspot/api.js
```diff
@@ -688,6 +688,28 @@ class Api extends OAuth2Requester {
...
+            url: `${this.baseUrl}${this.URLs.landingPages}/schedule`,
+            body: {
+                id: objId,
+                publishDate
+            },
+            headers: {
+                'Content-Type': 'application/json',
+                Accept: 'application/json',
+
+            }
```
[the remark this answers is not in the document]
hudsontavares: The suggested approach above would be backward compatible, I think.

api-module-library/hubspot/tests/api.test.js
MichaelRyanWebber: There's an underlying issue with the Requester class that the API inherits from that returns an empty string for successful post with no body in the response. I could modify the options to get the full response object, and check the status on that. I've done that previously, but I also don't like exposing the underlying node implementation (i.e. passing back the fetch response object).

I could also change the test to `expect(response).toBe("");` but that seems like going in the wrong direction / makes the test brittle to actually fixing the post handling in Requester. 

Because anything other than an underlying 200 status response will cause this to fail, I'd like to just keep it as is.

api-module-library/hubspot/tests/api.test.js
MichaelRyanWebber: See previous answer.

api-module-library/hubspot/api.js
MichaelRyanWebber: I agree that there are better ways to handle the URL formation. 

The current approach is limited by having this.URLs being defined in the constructor. I would like to change this so that it is a computed property, and whenever referenced the URLs are regenerated with current context. 

That's a little besides the point, what I really mean to get to is that there's no good reason that `this.URLs.landingPageById()` shouldn't already include the baseUrl, along with all other members of this.URLs

However, the convention in all other API modules is to use the baseURL as a prefix so I'm continuing that practice here. 

In working on the generator for API modules, we can define a new best practice and then look to refactor this module later.

api-module-library/hubspot/api.js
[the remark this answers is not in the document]
MichaelRyanWebber: We could override _post to apply these headers, or look to change the Requester class to default to add these headers. But, given that some _post requests, even just in this api module, do not take these headers, I like the explicit addition rather than needing to explicitly subtract them. I imagine sending them regardless would not cause issue most of the time.

Maybe it would make sense to add a new Requester method like _postJson() which added these headers?

api-module-library/hubspot/tests/api.test.js
```diff
@@ -299,36 +299,50 @@ describe(`${config.label} API tests`, () => {
...
-            delete pageToUpdate['archivedAt'];
+        it('update a Landing page (minimal patch)' , async () => {
             const response = await api.updateLandingPage(
-                pageToUpdate.id,
-                pageToUpdate,
+                sampleLandingPage.id,
+                {htmlTitle: `test Landing page ${Date.now()}`},
                 true );
             expect(response).toBeDefined();
         });
```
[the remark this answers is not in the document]
MichaelRyanWebber: I like this approach, although it might not make sense to always define expect output in the arrange block, from my view.

That said I think it would be out of scope to change all the tests to be like that for this PR

api-module-library/hubspot/api.js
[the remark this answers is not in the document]
hudsontavares: @MichaelRyanWebber - let's keep this as is for now and try to get back to it later.

api-module-library/hubspot/tests/api.test.js
seanspeaks: This specific request returns an empty response, is what I'm taking from this @MichaelRyanWebber. Is that accurate?

Side note, for future requests if we ever need/want the raw response, there's a helper boolean option of `returnFullRes` https://github.com/friggframework/frigg/blob/96d612cbe023fdf1694ec9ec1913296e125b1e57/packages/module-plugin/requester/requester.js#LL77C25-L77C26

That's mostly for more complex requests that involve things like grabbing Location headers for redirects or working with binary etc.

api-module-library/hubspot/api.js
```diff
@@ -721,6 +743,29 @@ class Api extends OAuth2Requester {
         return this._patch(options);
     }
 
+    async pushSitePageDraftToLive(objId){
```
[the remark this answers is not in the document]
seanspeaks: I've been hesitant to press "play" on JSDoc for years. (Just searched Slack and as far back as 2018 we were discussing whether to implement.)

It seems like folks can overdo it, underdo it, and just have wildly varying ways of doing it.

If we can agree on a consistent, probably minimal style and knowing _when_ to document via JSDocs and when not, I'm game.

[Clean Code](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882) and that group is pretty opposed to in-line documentation... but I think they're mostly in the camp of "your code should self document, and you should have as minimal comment as possible"... @MichaelRyanWebber I think you were reading the book too?

Anywho, that's my weigh in. I'm game to adopt, if we can set a baseline consistent style that everyone agrees on/can easily adopt and adhere to.

api-module-library/hubspot/tests/api.test.js
MichaelRyanWebber: @seanspeaks yes it's an empty response. For posts the Requester turns that into an empty string rather than a response object with an empty body but with a status and headers. Same with patch and put. I consider it to be a bug, but I'm not %100.

And yeah, that's what i was talking about in my comment, I could switch these requests to `returnFullRes: true` but to me that obfuscates things, since that full response is a node-fetch response that uses getters for some things, whereas Requester returns an object. 

So rather than encoding current behavior (by testing for "") or working around it by having each request like this grab the fullResponse, pull status from it etc, just leave as is with the thought to improve Requester's handling of this.

hudsontavares: @MichaelRyanWebber - as any non-200 will cause an error (I assume it would raise a rejected promise, right?), we could rewrite this test like this then:

```
const response = api.pushLandingPageDraftToLive(sampleLandingPage.id);

// If the processing fails, Jest will say "Received promise rejected instead of resolved" instead of some Javascript-based error that is usually harder to track down
await expect(response).resolves.toBe(/* any value you may be expecting to come from the output */);
```

It would give a more meaningful output in case of failure since you don't have a tangible output to check against.

hudsontavares (approved):
I am approving it. Please check if my last suggestion makes sense.

api-module-library/hubspot/tests/api.test.js
MichaelRyanWebber: https://linear.app/lefthook/issue/LEF-441/re-asses-requesters-handling-of-empty-response-body-on-postpatchput

to continue conversation

api-module-library/hubspot/api.js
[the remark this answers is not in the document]
MichaelRyanWebber: @hudsontavares  Sounds good, added
https://linear.app/lefthook/issue/LEF-442/re-assess-requests-handling-of-headers-on-postpatchput

api-module-library/hubspot/api.js
MichaelRyanWebber: @hudsontavares @seanspeaks 
added: https://linear.app/lefthook/issue/LEF-443/develop-best-practice-for-url-handling-in-api-modules

api-module-library/hubspot/api.js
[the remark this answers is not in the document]
hudsontavares: @seanspeaks - thanks for the input. Ok, let's not start a cosmic discussion about our code practices as part of this focused PR, then :) I am fine with keeping it as is.

I see some big fish like [Angular](https://github.com/angular/angular/blob/main/packages/core/src/render/api.ts) and [React](https://github.com/facebook/react/blob/c8deb5db660f3f6eeca30d29b4bd48d9e82b3792/packages/react/src/ReactElement.js#L128) using JSDoc in parts of their public API, which makes me wonder if that isn't useful.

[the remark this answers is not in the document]
MichaelRyanWebber: @seanspeaks I've more listened to talks by the author, and then discussions about his approach. 

I agree with the consistent and minimal approach, and personally would prefer options that:

- create a web presentation, even with examples a la the postman docs, sample data/response
- capture the more broad lessons of the underlying api which would help to communicate to the user, gotchas etc

and not having each function include a set of comments about it's inputs and outputs and side-effects. Especially with Typescript, I do hope this could be self-documenting, and if not, yeah probably the function is doing too much.

PR #191, opened 2023-07-04
Feature/lef 270 list organizations sites

roboli:
In order to keep things clean and working in the backend, I had to append the v2 to this module.

### DIFF
--- a/api-module-library/sharepoint/defaultConfig.json
+++ b/api-module-library/sharepoint/defaultConfig.json
@@ -1,5 +1,5 @@
 {
-    "name": "microsoft-sharepoint",
+    "name": "microsoft-sharepoint-v2",
     "label": "Microsoft SharePoint",
     "productUrl": "https://microsoft.com/sharepoint",
     "apiDocs": "https://developer.microsoft.com/en-us/graph/graph-explorer",
--- a/api-module-library/sharepoint/jest-setup.js
+++ b/api-module-library/sharepoint/jest-setup.js
@@ -5,7 +5,7 @@ const parsed = {
     SHAREPOINT_SCOPE: 'sharepoint_scope_test',
     SHAREPOINT_CLIENT_ID: 'sharepoint_client_id_test',
     SHAREPOINT_CLIENT_SECRET: 'sharepoint_client_secret_test',
-    SHAREPOINT_REDIRECT_URI: 'http://redirect_uri_test'
+    REDIRECT_URI: 'http://redirect_uri_test'
 };
 
 dotenv.populate(process.env, parsed);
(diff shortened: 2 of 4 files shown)


api-module-library/sharepoint/defaultConfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "microsoft-sharepoint",
+    "name": "microsoft-sharepoint-v2",
```
seanspeaks, line 2: @roboli I think we keep this as normal here and override it locally in the Frigg App. Anyone coming to this fresh shouldn't have to wonder what "v2" is all about. You'd override it in the "entities" directory, and either change the config variable or the GetName method.

api-module-library/sharepoint/jest-setup.js
```diff
@@ -2,10 +2,10 @@ const dotenv= require('dotenv');
 const { globalSetup } = require('@friggframework/test-environment');
 
 const parsed = {
-    SHAREPOINT_SCOPE: 'sharepoint_scope_test',
-    SHAREPOINT_CLIENT_ID: 'sharepoint_client_id_test',
-    SHAREPOINT_CLIENT_SECRET: 'sharepoint_client_secret_test',
-    SHAREPOINT_REDIRECT_URI: 'http://redirect_uri_test'
+    SHAREPOINT_V2_SCOPE: 'sharepoint_scope_test',
```
seanspeaks, line 5: Keep these as is, override locally as needed

api-module-library/sharepoint/manager.js
```diff
@@ -26,10 +26,10 @@ class Manager extends ModuleManager {
 
         // initializes the Api
         const sharepointParams = {
-            client_id: process.env.SHAREPOINT_CLIENT_ID,
-            client_secret: process.env.SHAREPOINT_CLIENT_SECRET,
-            redirect_uri: `${process.env.SHAREPOINT_REDIRECT_URI}/microsoft-sharepoint`,
-            scope: process.env.SHAREPOINT_SCOPE,
+            client_id: process.env.SHAREPOINT_V2_CLIENT_ID,
```
seanspeaks, line 29: Keep these the same here, override locally as needed

leofmds (changes requested):
@roboli We shouldn't change the name of this public integration due to an internal project. 
Instead, we should internalize what needs to be done to comply with the changed name (i.e. you get microsoft-sharepoint-v2 from the frontend, but for some reason you need to change it to microsoft-sharepoint in the backend).
The only thing I agree here is to change the SHAREPOINT_REDIRECT_URI to REDIRECT_URI. Leave it like that and rollback the rest.

PR #221, opened 2023-09-20
Add the team entity to Slack if it doesn't exists and return it in pr…

leofmds:
Automatically creates and returns the team entity in Slack, in case user scopes exist in the auth flow. 
If there is no user scopes provided, it just creates the team entity.

<details>
  <summary>📦 Published PR as canary version: <code>Canary Versions</code></summary>
  <br />

  ```bash
  npm install @friggframework/api-module-slack@0.2.2-canary.221.3ee830b.0
  # or 
  yarn add @friggframework/api-module-slack@0.2.2-canary.221.3ee830b.0
  ```
</details>

### DIFF
--- a/api-module-library/slack/api.js
+++ b/api-module-library/slack/api.js
@@ -21,7 +21,6 @@ class Api extends OAuth2Requester {
             `${process.env.REDIRECT_URI}/slack`
         );
         this.access_token = get(params, 'access_token', null);
-        this.team_access_token = get(params, 'access_token', null);
 
         this.URLs = {
             // Auth
@@ -82,7 +81,6 @@ class Api extends OAuth2Requester {
 
     async setTokens(params) {
         this.access_token = get(params, 'access_token');
-        this.team_access_token = get(params, 'access_token');
         this.refresh_token = get(params, 'refresh_token', null);
         const authedUser = get(params, 'authed_user', null);
         const accessExpiresIn = get(params, 'expires_in', null);
--- a/api-module-library/slack/api.js
+++ b/api-module-library/slack/api.js
@@ -3,6 +3,7 @@ const qs = require('qs');
 const { OAuth2Requester } = require('@friggframework/module-plugin');
 const { get } = require('@friggframework/assertions');
 const { FetchError } = require('@friggframework/errors');
+const moment = require("moment/moment");
 
 class Api extends OAuth2Requester {
     constructor(params) {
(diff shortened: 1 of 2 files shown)


api-module-library/slack/api.js
```diff
@@ -78,6 +80,28 @@ class Api extends OAuth2Requester {
...
+        if (authedUser && authedUser.access_token) {
+            this.access_token = authedUser.access_token;
+        }
+
+        this.accessTokenExpire = moment().add(accessExpiresIn, 'seconds');
+        this.refreshTokenExpire = moment().add(refreshExpiresIn, 'seconds');
+
+        await this.notify(this.DLGT_TOKEN_UPDATE);
+    }
+
```
MichaelRyanWebber, line 104: I'm not sure I fully understand this, but it seems like team_access_token is never really used? 
I think that it be better to override getTokenFromCode, since the job that setToken is doesn't need to change. The check around authedUser can be performed outside the scope of setTokens. 
Actually, maybe I understand that this is done also so that the refresh request does not need the same logic? In that case I might still slightly prefer leaving this lower level method with it's base implementation. 
But also, feel free to ignore.

api-module-library/slack/manager.js
```diff
@@ -150,7 +195,7 @@ class Manager extends ModuleManager {
 
     async updateOrCreateCredential() {
         const workspaceInfo = await this.api.authTest();
-        const externalId = get(workspaceInfo, 'team_id');
+        const externalId = workspaceInfo['bot_user_id'] ? get(workspaceInfo, 'team_id') : get(workspaceInfo, 'user_id');
```
MichaelRyanWebber, line 198: I'm guessing that this is determining whether we are updating a team or user credential? Might be worth a comment to indicate

MichaelRyanWebber (approved):
I wanted to suggest potentially renaming the findOrCreateEntity method to: findOrCreateUserEntity, if that is its exclusive role. 
I think I understand these changes, but it is a bit unclear. Seems to be creating both a user and team based entity, if both contexts are available. And if I have that right, then I think this makes sense.

api-module-library/slack/api.js
leofmds: I'm removing `team_access_token`. It was from a different implementation that I discarded later.

Here we are using the user access token if it exists, instead of the team token.

I think it makes more sense to update setTokens since the logic we're changing is exactly when setting the tokens. The rest stays the same.

api-module-library/slack/manager.js
leofmds: Changed the code to indicate, instead of adding a comment

seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-slack@0.2.3`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-slack%400.2.3) :rocket:

PR #230, opened 2023-12-05
Update Hubspot Credential model

Jo-lefthook:
Update Credential model columns to be snake_case
Update manager to use externalId on Credential instead of portalId Add setCredential function
Update redirect_uri

### DIFF
--- a/api-module-library/hubspot/manager.js
+++ b/api-module-library/hubspot/manager.js
@@ -31,8 +31,9 @@ class Manager extends ModuleManager {
             client_id: process.env.HUBSPOT_CLIENT_ID,
             client_secret: process.env.HUBSPOT_CLIENT_SECRET,
             scope: process.env.HUBSPOT_SCOPE,
-            redirect_uri: `${process.env.REDIRECT_URI}/hubspot`,
+            redirect_uri: process.env.REDIRECT_URI,
             delegate: instance,
+            state: JSON.stringify({ app: 'hubspot' })
         };
 
         if (params.entityId) {
@@ -41,8 +42,8 @@ class Manager extends ModuleManager {
                 instance.entity.credential
             );
             instance.credential = credential;
-            apiParams.access_token = credential.accessToken;
-            apiParams.refresh_token = credential.refreshToken;
+            apiParams.access_token = credential.access_token;
+            apiParams.refresh_token = credential.refresh_token;
         }
         instance.api = new Api(apiParams);
 
@@ -64,6 +65,12 @@ class Manager extends ModuleManager {
             url: await this.api.getAuthUri(),
             type: 'oauth2',
         };
+    }
+
+    setCredential(credential) {
+        this.credential = credential;
+        this.api.access_token = credential.access_token;
+        this.api.refresh_token = credential.refresh_token;
     }
 
     async processAuthorizationCallback(params) {
@@ -138,21 +145,22 @@ class Manager extends ModuleManager {
                 const userDetails = await this.api.getUserDetails();
                 const updatedToken = {
                     user: this.userId,
-                    accessToken: this.api.access_token,
-                    refreshToken: this.api.refresh_token,
-                    accessTokenExpire: this.api.accessTokenExpire,
-                    portalId: userDetails.portalId,
+                    access_token: this.api.access_token,
+                    refresh_token: this.api.refresh_token,
+                    access_token_expire: this.api.accessTokenExpire,
+                    expires_at: this.api.accessTokenExpire,
+                    externalId: userDetails.portalId,
                     auth_is_valid: true,
                 };
 
                 if (!this.credential) {
                     let credentialSearch = await Credential.find({
-                        portalId: userDetails.portalId,
+                        externalId: userDetails.portalId,
                     });
                     if (credentialSearch.length === 0) {
                         this.credential = await Credential.create(updatedToken);
                     } else if (credentialSearch.length === 1) {
-                        if (credentialSearch[0].user === this.userId) {
+                        if (credentialSearch[0].user.toString() === this.userId.toString()) {
                             this.credential = await Credential.findOneAndUpdate(
                                 { _id: credentialSearch[0] },
                                 { $set: updatedToken },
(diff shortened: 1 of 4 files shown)


api-module-library/hubspot/manager.js
```diff
@@ -31,8 +31,9 @@ class Manager extends ModuleManager {
             client_id: process.env.HUBSPOT_CLIENT_ID,
             client_secret: process.env.HUBSPOT_CLIENT_SECRET,
             scope: process.env.HUBSPOT_SCOPE,
-            redirect_uri: `${process.env.REDIRECT_URI}/hubspot`,
+            redirect_uri: process.env.REDIRECT_URI,
             delegate: instance,
+            state: JSON.stringify({ app: 'hubspot' })
```
seanspeaks, line 36: I think this one we want to pass in whatever state arg is provided by the caller for getAuthRequirements (this is not quite fully supported in concept yet but it's the logical place to add support) and remove the hardcoded json string here (it's a vestige of how FreshBooks' app was built)

(Then it's on us to pass it in on the get authorize route in our local Frigg application)

api-module-library/hubspot/manager.js
```diff
@@ -31,8 +31,9 @@ class Manager extends ModuleManager {
             client_id: process.env.HUBSPOT_CLIENT_ID,
             client_secret: process.env.HUBSPOT_CLIENT_SECRET,
             scope: process.env.HUBSPOT_SCOPE,
-            redirect_uri: `${process.env.REDIRECT_URI}/hubspot`,
+            redirect_uri: process.env.REDIRECT_URI,
```
seanspeaks, line 34: This one is tricky... @MichaelRyanWebber do you have an opinion on how we should navigate this? Each Frigg adopter may have a slightly different way they want to handle OAuth redirects, so if we're going to default to our "base url + api name" convention, we should have an easy override.

Overriding seems like a not so simple thing to do in Frigg v1 unless we have it at the module options level the integration definition?

In the current FreshBooks implementation, @Jo-lefthook, we should probably override the getInstance method in the local HubSpotEntityManager and change the redirect argument there instead of here in the published API Module

seanspeaks (changes requested):
See my notes. Some questions/answers may change a bit of my suggestions but you'll see my primary concern is on the state and redirect uri params. @MichaelRyanWebber's leanings for v1 will impact what I think we should put here for the pre-v1 release.

api-module-library/hubspot/manager.js
Jo-lefthook: @seanspeaks Yea, may we can start introducing an env variable for each integration library. for example, redirect_uri here would be something like `process.env.FRIGG_HUBSPOT_REDIRECT_URI || ${process.env.REDIRECT_URI}/hubspot`

api-module-library/hubspot/manager.js
Jo-lefthook: @seanspeaks The thing is, the `authorizationUri` is set in the [constructor](https://github.com/friggframework/frigg/blob/main/api-module-library/hubspot/api.js#L57) which prevents it from being updated later in the `getAuthRequirements` call. So we can either:
1- update the getAuthUri to read the state on the fly, similar to [slack](https://github.com/friggframework/frigg/blob/main/api-module-library/slack/api.js#L169)
2- accept the api state via the params of getInstance

Thoughts?

api-module-library/hubspot/manager.js
leofmds: @Jo-lefthook @seanspeaks  I think we should follow the same pattern as Microsoft Teams, check it out here: 
https://github.com/friggframework/frigg/blob/main/api-module-library/microsoft-teams/manager.js

It allows the user to override the redirect_uri, and if not it gets the default one `${process.env.REDIRECT_URI}/hubspot` already using Frigg's helpers

api-module-library/hubspot/manager.js
leofmds: @Jo-lefthook IMO the best way here is to change the `getAuthUri()` to re-generate the `authorizationUri` with the latest variable updates

api-module-library/hubspot/models/credential.js
leofmds, line 1: What's the reason to change the variable names? This would be a breaking change, so it should be updated only if really needed.

api-module-library/hubspot/manager.js
```diff
@@ -138,21 +145,22 @@ class Manager extends ModuleManager {
...
+                    access_token_expire: this.api.accessTokenExpire,
+                    expires_at: this.api.accessTokenExpire,
+                    externalId: userDetails.portalId,
                     auth_is_valid: true,
                 };
 
                 if (!this.credential) {
                     let credentialSearch = await Credential.find({
-                        portalId: userDetails.portalId,
+                        externalId: userDetails.portalId,
```
leofmds, line 158: This is also a breaking change but makes sense, so 👍🏻

api-module-library/hubspot/manager.js
```diff
@@ -41,8 +42,8 @@ class Manager extends ModuleManager {
                 instance.entity.credential
             );
             instance.credential = credential;
-            apiParams.access_token = credential.accessToken;
-            apiParams.refresh_token = credential.refreshToken;
+            apiParams.access_token = credential.access_token;
```
leofmds, line 45: Also related to a breaking change. I think we should start following a standardized naming convention for the names, instead of mixing camel case and snake case. I'm in favor of changing `apiParams.access_token` to `apiParams.accessToken` instead of changing the credential variables (see comment below).

api-module-library/hubspot/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@friggframework/api-module-hubspot",
-    "version": "0.9.1",
+    "version": "0.10.0",
```
leofmds, line 3: This PR contains breaking changes, the version should be bumped to `1.0.0` at least. Still, a question mark if this would be ready for a v1.

leofmds (changes requested):
I think we should at least clarify the naming changes before moving on.

api-module-library/hubspot/manager.js
Jo-lefthook: Cool, so we agree on the Microsoft Teams approach. I've pushed a commit implementing that. For the env variable, as @MichaelRyanWebber said, I think this should be added to the pre-release checklist to be implemented in all API modules

PR #247, opened 2024-01-18
Microsoft Teams - skip conversation reference re-creation

MichaelRyanWebber:
no longer re-create the conversation reference, by default, if a member already has one stored

### DIFF
--- a/api-module-library/microsoft-teams/api/api.js
+++ b/api-module-library/microsoft-teams/api/api.js
@@ -15,7 +15,7 @@ class Api {
         await this.botFrameworkApi.getTokenFromClientCredentials();
     }
 
-    async createConversationReferences(teamId=null){
+    async createConversationReferences(teamId=null, skipExisting=true){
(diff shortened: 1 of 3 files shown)


seanspeaks (approved):
Checks out

I think we don't have to manually bump the version, FYI, if we just add a tag to the PR of MINOR

MichaelRyanWebber:
> Checks out
> 
> I think we don't have to manually bump the version, FYI, if we just add a tag to the PR of MINOR

@MichaelRyanWebber yeah I thought so too (do we even need to tag? i think lerna might notice the change) but I didn't want to risk the back and forth this time so it was an extra precaution

seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-microsoft-teams@0.2.5`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-microsoft-teams%400.2.5) :rocket:

PR #249, opened 2024-02-03
Working concept for having a sane default list of environments where we bypass field-level encryption

seanspeaks:
@leofmds Thoughts on this approach?

I didn't get to run jest to tweak. Local error. Didn't feel like debugging.

Needed because now this defaults every env to encrypt, breaking folks inadvertently if they bump their packages (even any new api module update)

### DIFF
--- a/packages/encrypt/encrypt.js
+++ b/packages/encrypt/encrypt.js
@@ -15,11 +15,12 @@ const findOneEvents = [
 ];
 
 const shouldBypassEncryption = (STAGE) => {
-    if (!process.env.BYPASS_ENCRYPTION_STAGE) {
-        return false;
-    }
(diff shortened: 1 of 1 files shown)


leofmds (commented):
@seanspeaks  The idea I had in mind was to default to encryption, although we didn't think much about the previous implementations. This was added inadvertently as a patch change when it should be a major.
I'm in favor of reverting my changes or accepting your changes, and then adding the bypass as a breaking change. The motivation is that we would add the env variable only in cases where we explicitly want to bypass encryption. With the proposed changes, we still need to add a new env with any content in it to work in the use case that motivated my initial changes.
Let me know your thoughts on this.

PR #250, opened 2024-02-04
[Api Module] Fix: Google Calendar

nicolasmelo1:
This PR is for fixing the Google Calendar API module.

We are exporting the Definition through `index.js`
and that would be the first change.

The second change was a bug i have encountered on the Auther that was giving me errors.

@sean showed me that the fix to this error would be:
```js
apiPropertiesToPersist: {
  credential: ['access_token','refresh_token'],
  entity: []
},
```

Because [Auther expects](https://github.com/friggframework/frigg/blob/813908d9b9c52c1f67d7f02a205f74cb31554dd4/packages/module-plugin/auther.js#L69) `definition.requiredAuthMethods.apiPropertiesToPersist.credential` to exist.

This is a change specifically for v1-alpha, not for Main at the given moment.

### DIFF
--- a/api-module-library/google-calendar/definition.js
+++ b/api-module-library/google-calendar/definition.js
@@ -19,7 +19,7 @@ const Definition = {
         getEntityDetails: async function (api, callbackParams, tokenResponse, userId) {
             const entityDetails = await api.getTokenIdentity();
             return {
-                identifiers: { externalId: entityDetails.identifier, userId },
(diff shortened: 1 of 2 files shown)


seanspeaks:
:rocket: PR was released in [`v1.0.1-v1-alpha.0`](https://github.com/friggframework/frigg/releases/tag/v1.0.1-v1-alpha.0) :rocket:

PR #264, opened 2024-03-01
package update

MichaelRyanWebber:
# Version

Published prerelease version: `v1.0.1-v1-alpha-package-update.1`

<details>
  <summary>Changelog</summary>

  #### ⚠️ Pushed to `v1-alpha-package-update`
  
  - `@friggframework/api-module-42matters`
    - bump just 42matters as a test ([@MichaelRyanWebber](https://github.com/MichaelRyanWebber))
  - `@friggframework/api-module-42matters`, `@friggframework/api-module-activecampaign`, `@friggframework/api-module-airwallex`, `@friggframework/api-module-attentive`, `@friggframework/api-module-clyde`, `@friggframework/api-module-connectwise`, `@friggframework/api-module-contentful`, `@friggframework/api-module-contentstack`, `@friggframework/api-module-crossbeam`, `@friggframework/api-module-deel`, `@friggframework/api-module-fastspring-iq`, `@friggframework/api-module-freshbooks`, `@friggframework/api-module-front`, `@friggframework/api-module-frontify`, `@friggframework/api-module-google-calendar`, `@friggframework/api-module-google-drive`, `@friggframework/api-module-gorgias`, `@friggframework/api-module-hubspot`, `@friggframework/api-module-huggg`, `@friggframework/api-module-ironclad`, `@friggframework/api-module-linear`, `@friggframework/api-module-marketo`, `@friggframework/api-module-microsoft-teams`, `@friggframework/api-module-monday`, `@friggframework/api-module-netx`, `@friggframework/api-module-outreach`, `@friggframework/api-module-personio`, `@friggframework/api-module-pipedrive`, `@friggframework/api-module-qbo`, `@friggframework/api-module-rev-io`, `@friggframework/api-module-rollworks`, `@friggframework/api-module-salesforce`, `@friggframework/api-module-salesloft`, `@friggframework/api-module-microsoft-sharepoint`, `@friggframework/api-module-slack`, `@friggframework/api-module-terminus`, `@friggframework/api-module-unbabel`, `@friggframework/api-module-yotpo`, `@friggframework/api-module-zoom`, `@friggframework/core`, `@friggframework/devtools`
    - specify fixed version correctly ([@MichaelRyanWebber](https://github.com/MichaelRyanWebber))
  
  #### Authors: 1
  
  - [@MichaelRyanWebber](https://github.com/MichaelRyanWebber)
</details>


api-module-library/42matters/package.json
```diff
@@ -1,31 +1,28 @@
...
-  "publishConfig": {
-    "access": "public"
-  },
-  "dependencies": {
-    "md5": "^2.3.0"
-  }
-}
+    "name": "@friggframework/api-module-42matters",
+    "version": "0.0.5",
+    "prettier": "@friggframework/prettier-config",
```
seanspeaks, line 4: Is this now `@friggframework/devtools/prettier-config` or should we release a prettier-config?

api-module-library/42matters/package.json
```diff
@@ -1,31 +1,28 @@
...
+        "jest": "^29.7.0",
+        "prettier": "^3.0.3",
+        "sinon": "^16.0.0"
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "dependencies": {
+        "md5": "^2.3.0",
+        "@friggframework/core": "^1.0.0"
```
seanspeaks, line 26: Per Andrew's comment in Discord, should this be a local/workspace reference to let Lerna do the crawling? or are we intentionally version pinning?

api-module-library/42matters/package.json
```diff
@@ -1,31 +1,28 @@
...
+    "description": "",
+    "main": "index.js",
+    "scripts": {
+        "lint:fix": "prettier --write --loglevel error . && eslint . --fix",
+        "test": "jest"
+    },
+    "author": "",
+    "license": "MIT",
+    "devDependencies": {
+        "@friggframework/devtools": "^1.0.0",
```
seanspeaks, line 14: Ditto here for my other point

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "useNx": true,
-  "version": "independent"
+  "version": "1.0.0"
```
seanspeaks, line 4: Ah is this related to the Andrew comment/suggestion? What does this do for us?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "frigg-monorepo",
-  "version": "1.0.4",
+  "version": "1.0.0",
```
seanspeaks, line 3: Does down-bumping this matter at all? I think the answer is "no", because this isn't actually packaging/releasing, right? This could stay unversioned?

package.json
```diff
@@ -22,22 +22,21 @@
...
-    "@auto-it/slack": "^11.0.4",
-    "auto": "^11.0.4",
+    "@auto-it/all-contributors": "^11.1.1",
+    "@auto-it/conventional-commits": "^11.1.1",
+    "@auto-it/first-time-contributor": "^11.1.1",
+    "@auto-it/slack": "^11.1.1",
+    "auto": "^11.1.1",
     "lerna": "^7.4.2",
-    "nx": "^17.2.6"
+    "nx": "^16.10.0"
```
seanspeaks, line 31: Oh interesting, is nx 17 not compatible?

packages/core/module-plugin/requester/oauth-2.js
```diff
@@ -44,8 +43,8 @@ class OAuth2Requester extends Requester {
             null
         );
 
-        this.accessTokenExpire = moment().add(accessExpiresIn, 'seconds');
-        this.refreshTokenExpire = moment().add(refreshExpiresIn, 'seconds');
+        this.accessTokenExpire = new Date(Date.now() + accessExpiresIn * 1000);
```
seanspeaks, line 46: This just a "moment too heavy" thing? or this wasn't working?

packages/core/module-plugin/test/mock-api/api.js
```diff
@@ -1,5 +1,5 @@
-const { get } = require('@friggframework/assertions');
-const { OAuth2Requester } = require('@friggframework/module-plugin');
+const { get } = require('../../assertions');
```
seanspeaks, line 1: Does this matter if we use local reference vs the package? (Wondering if in some cases it would be better to reference the package/workspace, vs the path, since pathing may change as we iterate)

packages/core/package.json
```diff
@@ -1,12 +1,35 @@
 {
   "name": "@friggframework/core",
   "prettier": "@friggframework/prettier-config",
-  "version": "0.2.30",
+  "version": "1.0.0",
```
seanspeaks, line 4: Is this on us to do, or should this be a tag that we add to the PR into `main`, with a `major` tag so that auto/lerna does its magic?

packages/core/package.json
```diff
@@ -1,12 +1,35 @@
...
+  "dependencies": {
+    "bcryptjs": "^2.4.3",
+    "mongoose": "6.11.6",
+    "jest-runner-groups": "^2.2.0",
+    "mongodb-memory-server": "^8.9.0",
+    "lodash": "^4.17.21",
+    "lodash.get": "^4.4.2",
+    "aws-sdk": "^2.1200.0",
```
seanspeaks, line 12: Can this go to dev?

packages/core/package.json
```diff
@@ -1,12 +1,35 @@
 {
   "name": "@friggframework/core",
   "prettier": "@friggframework/prettier-config",
-  "version": "0.2.30",
+  "version": "1.0.0",
+  "dependencies": {
+    "bcryptjs": "^2.4.3",
+    "mongoose": "6.11.6",
+    "jest-runner-groups": "^2.2.0",
+    "mongodb-memory-server": "^8.9.0",
```
seanspeaks, line 9: This and jest-runner-groups should go to devDependencies, yeah?

packages/devtools/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@friggframework/devtools",
+  "prettier": "@friggframework/prettier-config",
+  "version": "1.0.0",
+  "dependencies": {
+    "@friggframework/core": "^1.0.0",
```
seanspeaks, line 6: This may be semantic, but in theory couldn't all of these be devDependencies?

seanspeaks (approved):
Approving, since most of my comments are curiosity/informative. They may result in later changes but I don't think it should hold up a merge.

Lots of work represented here 🙏 Thanks!

lerna.json
MichaelRyanWebber: this change didn't seem to do anything for us, that I could tell.

package.json
MichaelRyanWebber: yes I don't think it should matter, though I'm not sure... but it seems like a good idea try to release on 1.0.0

api-module-library/42matters/package.json
MichaelRyanWebber: I'm happy to switch to that, I was going to try to find some examples before updating. 

What do you think? That would be like:

`"@friggframework/core": "packages:core"` 

or more literally:

`"@friggframework/core": "workspace:*"`

api-module-library/42matters/package.json
MichaelRyanWebber: oh hmm, interesting. i don't know if `@friggframework/devtools/prettier-config` would work, since it's exported at the root of devtools and isn't a separate package
`@friggframework/devtools` would be enough here?

packages/core/package.json
MichaelRyanWebber: shrug, I think there are many ways we could do it. I don't know what we'd be gaining by trying to get lerna to do something

packages/core/package.json
MichaelRyanWebber: the memory server? yes

packages/core/package.json
MichaelRyanWebber: yes

packages/devtools/package.json
MichaelRyanWebber: yeah I had that thought too, wasn't sure what to make of that. they could also all be just regular dependencies since the dev is implied

packages/core/module-plugin/test/mock-api/api.js
MichaelRyanWebber: I know that it matters in some cases because it can create circular imports and such. I think if it's in the same package it should use the local reference, but maybe I'm wrong on that

package.json
MichaelRyanWebber: yes it's not recommended with that lerna version, which is the highest that auto explicitly supports

api-module-library/42matters/package.json
MichaelRyanWebber: @seanspeaks this one is the most important I think. But I won't wait on merge

seanspeaks: Worth a look probably module by module for our desired behavior as we iterate through stuff.

Tests will help. Then we can just run tests as we make changes and feel somewhat confident we didn't break everything 😂

PR #276, opened 2024-03-19
Release Frigg v1

MichaelRyanWebber:


api-module-library/42matters/bump0.txt
seanspeaks, line 1: Delete?

docs/README.md
seanspeaks, line 1: Midlly worried about what this will do to the docs. But not reason to not merge. Old docs are fine to fully lose at hte moment 😂

seanspeaks: Scratch that. I disconnected the integration, so we'll hook it back up once we're clear.

seanspeaks (approved):
Let's go 👍

PR #280, opened 2024-03-20
Fix MS Teams refresh token and patch version 0.

brendanrmoore:
Currently, the refresh token is not getting passed in when instantiating the graph API, so this prevents tokens from refreshing. This was actually fixed recently in version 1, but I would like to apply this fix on version 0 as well.

### DIFF
--- a/api-module-library/microsoft-teams/package.json
+++ b/api-module-library/microsoft-teams/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@friggframework/api-module-microsoft-teams",
-    "version": "0.4.1",
+    "version": "0.4.0",
     "prettier": "@friggframework/prettier-config",
     "description": "",
     "main": "index.js",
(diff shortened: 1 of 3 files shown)


api-module-library/microsoft-teams/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@friggframework/api-module-microsoft-teams",
-    "version": "0.4.0",
+    "version": "0.4.1",
```
MichaelRyanWebber, line 3: @brendanrmoore we actually don't need to bump the version! Although we have done plenty of that lately. It should default to consider this a patch update, but we can also label the PR.

MichaelRyanWebber (approved):
Thanks for this! 
Undoing the version bump isn't strictly necessary

api-module-library/microsoft-teams/package.json
brendanrmoore: @MichaelRyanWebber ahh, I see. I reverted that commit anyway to clean up the PR. If it looks good, do you mind merging it in? Thanks for reviewing!

seanspeaks:
:rocket: PR was released in [`@friggframework/api-module-microsoft-teams@0.4.1`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/api-module-microsoft-teams%400.4.1) :rocket:

PR #291, opened 2024-03-28
Implement createFile function in google drive

Jo-lefthook:
This PR adds a `createFile` function to the google-drive api module

### DIFF
--- a/api-module-library/google-drive/package.json
+++ b/api-module-library/google-drive/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@friggframework/api-module-google-drive",
-  "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


api-module-library/google-drive/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@friggframework/api-module-google-drive",
-  "version": "1.1.1",
+  "version": "1.1.2",
```
leofmds, line 3: You can remove this version bump as it will be done automatically through the `patch` label.

leofmds (changes requested):
Required one small change

leofmds (approved):
👍🏻

PR #297, opened 2024-04-03
Socket hang up / ECONNRESET error retry for requester

seanspeaks:
We've been running into a socket hang up error during Square API requests. This was one snippet of code that seemed to help; I'm not sure it's the most efficient retry logic?

Only other thought... we're going to have to manually bump the API module we care about once this gets merged/released, yeah? Now I'm remembering one benefit of lerna/auto... though we can/should have a chore that just dependabot bumps dependencies in api-module-library.

### DIFF
--- a/packages/core/module-plugin/requester/requester.js
+++ b/packages/core/module-plugin/requester/requester.js
@@ -48,7 +48,21 @@ class Requester extends Delegate {
 
         if (this.agent) options.agent = this.agent;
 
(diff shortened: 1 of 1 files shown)


MichaelRyanWebber (approved):
LGTM

leofmds (approved):
This improvement makes sense to me 👍🏻

seanspeaks:
Merging but going to do a second PR in version-0... fully diverging from each other

PR #298, opened 2024-04-03
Removed console.log for Slack responses

leofmds:


api-module-library/slack/api.js
```diff
@@ -117,7 +117,6 @@ class Api extends OAuth2Requester {
         const parsedResponse = await this.parsedBody(response);
         const { status } = response;
         const { ok, error } = parsedResponse;
-        console.log(parsedResponse);
```
seanspeaks, line 120: Mm yeah. Though I think this needs to be based on version-0 and PR'ed into version-0?

@leofmds cc @MichaelRyanWebber

leofmds: I'll do it to version-0 and then v1. Should we still use this repo to update the modules for v1?
@seanspeaks

seanspeaks: I'm 90% sure that any version over 0.x.x goes to api-module-library.

So, this would be a PR into version-0 and a PR into `main` over at api-module-library

seanspeaks: @leofmds / @MichaelRyanWebber Looping back on this before EOD (I see that you changed the branch back to `main` after toggling to `version-0` @leofmds )

PR #304, opened 2024-04-15
small update to integration testing / tooling

MichaelRyanWebber:


seanspeaks (approved):
Neat- 

We'll probably need a current "as is" state write-up with a "to be" state... for example, seeing that our base test assumes entity 1 and entity 2, which will need to be updated once we move to named modules, etc.

Do you have a working codebase where this specific update is relevant? (I think the answer is yes, would love to review at some point)

seanspeaks:
:rocket: PR was released in [`v1.2.0`](https://github.com/friggframework/frigg/releases/tag/v1.2.0) :rocket:

PR #305, opened 2024-04-18
Remove user ID check from createIntegration

brendanrmoore:
The Slack API module uses a many to 1 mapping of integrations to Slack entities when those integrations all connect to the same Slack workspace. However, this conflicts with the user ID check here which will throw an error if the user IDs don't match.

### DIFF
--- a/packages/integrations/manager.js
+++ b/packages/integrations/manager.js
@@ -180,12 +180,6 @@ class IntegrationManager extends Delegate {
             if (!entity) {
                 throw new Error(`Entity with ID ${id} does not exist.`);
             }
-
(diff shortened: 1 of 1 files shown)


MichaelRyanWebber (approved):
Looks good to me! This has been running (without this check) in v1 with no problem.

seanspeaks:
:rocket: PR was released in [`@friggframework/integrations@1.2.5`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/integrations%401.2.5), [`@friggframework/types@0.2.5`](https://github.com/friggframework/frigg/releases/tag/%40friggframework/types%400.2.5) :rocket:

PR #313, opened 2024-05-27
Fixes for Frigg CI

igorschechtel:
The 'Frigg CI' workflow is currently failing, this PR is supposed to fix the errors

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6629,20 +6629,6 @@
         "semver": "^7.0.0"
       }
     },
-    "node_modules/bundle-name": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/bundle-name/-/bundle-name-4.1.0.tgz",
-      "integrity": "sha512-tjwM5exMg6BGRI+kNmTntNsvdZS1X8BFYS6tnJ2hdH0kVxM6/eVZ2xy+FqStSWvYmtfFMDLIxurorHwDKfDz5Q==",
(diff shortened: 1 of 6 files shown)


igorschechtel:
Hey @seanspeaks or @MichaelRyanWebber, I'm trying to make the CI pass. Do you have any idea why this test is failing?

```
FAIL module-plugin/test/auther.test.js
  ● HubSpot Module Tests › Test credential retrieval and module instantiation › retrieve by entity id

    TypeError: Cannot read properties of null (reading 'credential')

      113 |                 params.credentialId
      114 |             );
    > 115 |         }
          |          ^
      116 |         let credential = {};
      117 |         let entity = {};
      118 |         if (instance.credential) {

```

MichaelRyanWebber:
@igorschechtel I wonder if perhaps the in-memory mongo is not working in CI. It's passing for me when I run it (thanks to your fixes to the imports), is that the same for you?

igorschechtel:
@MichaelRyanWebber yes, this test is also passing when I run it locally.

It seems to me that the issue is that `instance` is `null`. So when it tries to access `instance.credential` we get the error:
```
TypeError: Cannot read properties of null (reading 'credential')
```

If this is true I think the origin for the error is line 105:

```
const instance = new this(params);
```

Not sure why this would return null though... maybe an error in the constructor? The fact that it passes locally but not in the workflow makes it even weirder.

PR #322, opened 2024-08-01
CLI for Frigg - Install command for now

seanspeaks:
New CLI for Frigg, with the first command being `frigg install <apiModuleName>`. We'll see where it takes us.

Some ideas/areas of improvement: 
- [ ] tests 
- [ ] more prompts 
- [ ] aws logs interfering 
- [ ] more robust colors etc, maybe use some tooling
- [ ] detect where envs are handled and just add there instead of both places
- [ ] make the envs retain their caps case with the app name prefix 
- [ ] Add prompts for editing the integration concept 
- [ ] Add more commands that would be helpful during development


seanspeaks:
Task linked: [FRI-418 Explore ability to have a CLI as part of devtools package](https://app.clickup.com/t/9013279251/FRI-418)

packages/devtools/frigg-cli/backendPath.js
```diff
@@ -0,0 +1,26 @@
...
+function findNearestBackendPackageJson() {
+    let currentDir = process.cwd();
+    while (currentDir !== path.parse(currentDir).root) {
+        const packageJsonPath = path.join(currentDir, 'backend', PACKAGE_JSON);
+        if (fs.existsSync(packageJsonPath)) {
+            return packageJsonPath;
+        }
+        currentDir = path.dirname(currentDir);
+    }
+    return null;
```
d-klotz, line 14: should we return null? I would expect to see an error (eg throwed) if no package.json is found.

packages/devtools/frigg-cli/environmentVariables.js
```diff
@@ -0,0 +1,134 @@
...
+                            exp.type === 'MemberExpression'
+                                ? `${exp.property.name}`
+                                : exp.name
+                        );
+                        envVariables[key] = expressions.join('');
+                    }
+                });
+            }
+        },
+    });
```
d-klotz, line 42: Did you copy that from somewhere ? 😂
Hard to follow

d-klotz (approved):

packages/devtools/frigg-cli/backendPath.js
seanspeaks: I believe running npm install on the backend path will initialize and generate? Worth testing at some point, but also expecting that this will almost never run into this logic unless someone gets creative with their frigg app setup

packages/devtools/frigg-cli/environmentVariables.js
seanspeaks: It was a generated code snippet. Can refactor later for comprehension 👍

seanspeaks:
:rocket: PR was released in [`v1.2.0`](https://github.com/friggframework/frigg/releases/tag/v1.2.0) :rocket:

PR #326, opened 2024-08-08
Fix bug during local running

seanspeaks:
When running an app locally, the first time an integration record is retrieved by ID and the module-factory is run, the app errors out.

Specifically, `getModuleInstanceFromEntityId` fails to find the moduleDefinition, only on the first invocation.

After debugging with logs and inspection, it looks like mongoose (current version?) fails to initialize the descriminator decorator on the mongoose object being evaluated (`entity['__t']` comes up `undefined`). Subsequent requests succeed. This makes me believe there's something to do with initializing the mongoDB connection or mongoose collections. 

One patch fix here is to force the mongoose object to finish initializing by calling toJSON.

Works locally for me, unclear how to write the test to prove it out.

### DIFF
--- a/packages/core/module-plugin/module-factory.js
+++ b/packages/core/module-plugin/module-factory.js
@@ -1,12 +1,10 @@
-const { Entity } = require("./entity");
+const { Entity } = require('./entity');
 const { Auther } = require('./auther');
 
 class ModuleFactory {
     constructor(...params) {
         this.moduleDefinitions = params;
-        this.moduleTypes = this.moduleDefinitions.map(
-            (def) => def.moduleName
-        );
+        this.moduleTypes = this.moduleDefinitions.map((def) => def.moduleName);
     }
 
     async getEntitiesForUser(userId) {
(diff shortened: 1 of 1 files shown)


d-klotz (approved):
We should create a task to investigate further, this PR fixes the issue but the root cause is still not clear and can probably happen to other methods in there, right?

seanspeaks:
:rocket: PR was released in [`v1.2.1`](https://github.com/friggframework/frigg/releases/tag/v1.2.1) :rocket:

seanspeaks:
> We should create a task to investigate further, this PR fixes the issue but the root cause is still not clear and can probably happen to other methods in there, right?

So created 👍 FRI-459

MichaelRyanWebber:
@seanspeaks and @d-klotz You're definitely on to something there Sean. 

This is the same issue that I had to work around in getInstance:

https://github.com/friggframework/frigg/blob/main/packages/core/module-plugin/auther.js#L119

I also remember seeing this behavior before, I had added the error check but hadn't been seeing the error, so hadn't put it together. 

Will be great to get to the root cause

PR #327, opened 2024-08-09
Add support for secrets loading from SECRET_ARN

seanspeaks:
AWS specific that could be modified to be independent

Will not attempt if no SECRET_ARN is set in ENVs

### DIFF
--- a/packages/core/core/create-handler.js
+++ b/packages/core/core/create-handler.js
@@ -3,6 +3,7 @@ require('source-map-support').install();
 
 const { connectToDatabase } = require('../database/mongo');
 const { initDebugLog, flushDebugLog } = require('../logs');
+const { secretsToEnv } = require('./secrets-to-env');
 
 const createHandler = (optionByName = {}) => {
     const {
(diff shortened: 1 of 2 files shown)


seanspeaks:
Task linked: [FRI-460 Move SecretsToEnv to frigg core, add to createHandler](https://app.clickup.com/t/9013279251/FRI-460)

packages/core/core/create-handler.js
```diff
@@ -20,6 +21,15 @@ const createHandler = (optionByName = {}) => {
         try {
             initDebugLog(eventName, event);
 
+            const requestMethod = event.httpMethod;
+            const requestPath = event.path;
+            if (requestMethod && requestPath) {
```
d-klotz, line 26: what's the use of this variables besides the loging?

seanspeaks: Just logging

(Thought it was useful context in another project... could remove or really deserves a full revamp of logging/telemetry across Frigg)

packages/core/core/secrets-to-env.js
```diff
@@ -0,0 +1,61 @@
+const getSecretValue = async () => {
+    console.log('Fetching secrets...');
+
+    const httpPort = process.env.PARAMETERS_SECRETS_EXTENSION_HTTP_PORT || 2773;
```
d-klotz, line 4: Should we really default here? If this is a env var used in more places, we should throw an error if it's not provided.

d-klotz (approved):

seanspeaks:
:rocket: PR was released in [`v1.2.2`](https://github.com/friggframework/frigg/releases/tag/v1.2.2) :rocket:

PR #429, opened 2025-09-22
Fix: Resolve KMS VPC Endpoint Connectivity Timeout Issue

seanspeaks:
## Summary
- Fixed KMS VPC endpoint connectivity timeout issue by correcting security group ingress rules
- Lambda functions can now successfully connect to KMS through VPC endpoints
- Resolves TCP port 443 connection timeouts seen in production

## Problem
The VPC endpoint's security group was only allowing inbound traffic from the VPC CIDR block, not specifically from the Lambda function's security group. This caused connection timeouts when Lambda tried to access KMS through the VPC endpoint.

## Solution
Enhanced the VPC endpoint security group configuration to:
1. **Prioritize Lambda security group** - Allow traffic specifically from Lambda's security group (most secure)
2. **Fallback to VPC CIDR** - Use VPC CIDR if Lambda SG not available
3. **Default fallback** - Use common private IP ranges as last resort
4. **Support both scenarios** - Works for both create-new and discover VPC modes

## Changes
- Modified `serverless-template.js` to create proper security group ingress rules
- Added intelligent rule selection based on available resources
- Enhanced logging and debugging capabilities
- Added descriptive tags for better resource management

## Test Results
From the logs provided, we can see:
- ✅ DNS resolution works: KMS endpoint resolves to VPC endpoint IPs (`172.31.42.140`, `172.31.6.38`)
- ❌ TCP connection failed: Connection timeout on port 443 (before fix)
- ✅ After fix: Security group will allow Lambda → VPC Endpoint traffic

## Deployment Instructions
1. Deploy this change to update CloudFormation stack
2. The security group rules will be automatically updated
3. Lambda functions will be able to connect to KMS without timeouts

## Verification
After deployment, the health check should show:
- KMS connectivity: `healthy`
- No more "Connection timeout" errors
- Successful encryption/decryption operations

### DIFF
--- a/packages/devtools/infrastructure/serverless-template.js
+++ b/packages/devtools/infrastructure/serverless-template.js
@@ -670,7 +670,7 @@ const createVPCInfrastructure = (AppDefinition) => {
         vpcResources.FriggVPCEndpointSecurityGroup = {
             Type: 'AWS::EC2::SecurityGroup',
             Properties: {
-                GroupDescription: 'Security group for Frigg VPC Endpoints',
+                GroupDescription: 'Security group for Frigg VPC Endpoints - allows HTTPS from Lambda functions',
                 VpcId: { Ref: 'FriggVPC' },
                 SecurityGroupIngress: [
                     {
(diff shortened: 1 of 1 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.40`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.40) :rocket:

PR #462, opened 2025-10-16
Adds support for default webhook handling, including integration instance hydration for the handlers. Adds two new default events Adds tests (Docs can be removed)

seanspeaks:
# Webhook Handling Implementation

Adds comprehensive webhook support for Frigg integrations with a scalable architecture that separates HTTP response from async processing.

## Key Features

- Added two new lifecycle events to `IntegrationBase`:
  - `WEBHOOK_RECEIVED` - Handles HTTP requests without database connection
  - `ON_WEBHOOK` - Processes webhooks with full integration context

- Implemented default handlers:
  - `onWebhookReceived()` - Queues webhooks to SQS and responds quickly
  - `onWebhook()` - Processes webhooks with database access
  - `queueWebhook()` - Utility to send messages to SQS

- Added webhook HTTP router that creates routes:
  - `POST /api/{name}-integration/webhooks`
  - `POST /api/{name}-integration/webhooks/:integrationId`

- Enhanced queue worker to support hydrated integration instances

- Updated serverless template to automatically generate webhook functions

- Added comprehensive test coverage:
  - Unit tests for event dispatching, router configuration, and worker processing
  - Integration tests for end-to-end webhook flow

### DIFF
--- a/WEBHOOK_IMPLEMENTATION_SUMMARY.md
+++ b/WEBHOOK_IMPLEMENTATION_SUMMARY.md
@@ -1,0 +1,256 @@
+# Webhook Handling Implementation Summary
+
+## Overview
+
+Implemented comprehensive webhook handling support for Frigg integrations with a scalable, database-efficient architecture that separates HTTP response from async processing.
+
+## What Changed
+
+### Core Infrastructure
+
+#### 1. Event Constants & Default Handlers
+**File:** `packages/core/integrations/integration-base.js`
+
+Added two new lifecycle events:
+- `WEBHOOK_RECEIVED` - HTTP handler (no database connection)
+- `ON_WEBHOOK` - Queue worker (database-connected, hydrated instance)
+
+```javascript
+// New constants
+constantsToBeMigrated.defaultEvents.WEBHOOK_RECEIVED
+constantsToBeMigrated.defaultEvents.ON_WEBHOOK
+
+// Default handlers
+async onWebhookReceived({ req, res }) { /* queue to SQS */ }
+async onWebhook({ data }) { /* process with full context */ }
+async queueWebhook(data) { /* send to SQS */ }
+```
+
+#### 2. Queue Utilities
+**File:** `packages/core/queues/queuer-util.js`
+
+Added single message send method:
+```javascript
+QueuerUtil.send(message, queueUrl)
+```
+
+#### 3. Webhook HTTP Router
+**File:** `packages/core/handlers/routers/integration-webhook-routers.js` (NEW)
+
+Creates webhook routes for integrations with `webhooks: true`:
+- `POST /api/{name}-integration/webhooks` - General webhooks
+- `POST /api/{name}-integration/webhooks/:integrationId` - Integration-specific
+
+#### 4. Enhanced Queue Worker
+**File:** `packages/core/handlers/backend-utils.js`
+
+Updated `createQueueWorker` to support hydrated instances:
+- Detects `ON_WEBHOOK` events with `integrationId`
+- Loads full integration from database
+- Provides access to API modules and integration context
+
+#### 5. Serverless Infrastructure
+**File:** `packages/devtools/infrastructure/serverless-template.js`
+
+Automatically generates webhook Lambda functions when `webhooks: true`:
+```yaml
+functions:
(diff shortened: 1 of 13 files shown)


seanspeaks:
* **#462** <a href="https://app.graphite.dev/github/pr/friggframework/frigg/462?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/friggframework/frigg/462?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `next`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

WEBHOOK_IMPLEMENTATION_SUMMARY.md
```diff
@@ -0,0 +1,256 @@
...
+
+```javascript
+class MyIntegration extends IntegrationBase {
+    static Definition = {
+        name: 'my-integration',
+        version: '1.0.0',
+        modules: {
+            myapi: { definition: MyApiDefinition },
+        },
+        webhooks: true, // Enable webhooks
```
d-klotz, line 94: Why would we NOT use webhooks?

packages/core/handlers/backend-utils.js
```diff
@@ -37,7 +37,42 @@ const createQueueWorker = (integrationClass) => {
     class QueueWorker extends Worker {
         async _run(params, context) {
             try {
-                const integrationInstance = new integrationClass();
+                let integrationInstance;
+
+                // Check if this is a webhook with integrationId
+                if (params.event === 'ON_WEBHOOK' && params.data?.integrationId) {
+                    // Load hydrated integration instance
+                    const { GetIntegrationInstance } = require('../integrations/use-cases/get-integration-instance');
```
d-klotz, line 45: All this code will go away in the future when we introduce a decent IoC container.

WEBHOOK_IMPLEMENTATION_SUMMARY.md
d-klotz, line 94: Also, it feels like this should be in the appDefintion rather than here in integration Definition, no ?

d-klotz (approved):

WEBHOOK_IMPLEMENTATION_SUMMARY.md
d-klotz, line 94: And I'd probably be happier if we used `enableWebhooks` instead of just `webhooks`.

seanspeaks: Extra bloat/cognitive load. If you have 10 integrations, suddenly our number of lambdas and logs etc. goes from 10 to 20.

seanspeaks: Per integration; some integrations don't need it. appDefinition could be an optional blanket instead of turning it on per integration though.

seanspeaks: I wanted to do an optional "boolean for default logic" and then an object for future extensibility, so you'll see in the logic that we have a check for `webhooks` and `webhooks.enabled`

The idea being that we could eventually do something like
```
webhooks: {
    enabled: true,
    
    // Future properties we could support:
    path: '/custom-webhook-path',  // Custom endpoint suffix
    signatureHeader: 'x-custom-signature',  // Which header to check
    signatureAlgorithm: 'sha256',  // or 'sha1', 'sha512'
    timeout: 5000,  // HTTP timeout in ms
    batchSize: 10,  // Process N webhooks at once in worker
    retryAttempts: 3,  // How many times to retry failed processing
    filterEvents: ['item.created', 'item.updated'],  // Only queue these events
    queueDelay: 0,  // Delay before worker processes (in seconds)
}
```

So, declarative extensions instead of making someone build the code to do each of those things. (Some of those don't make sense on the face, but you get the idea)

seanspeaks:
:rocket: PR was released in [`v2.0.0-next.45`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.45) :rocket:

PR #468, opened 2025-10-23
fix(infrastructure): Discover KMS key alias even if not in CloudForma…

seanspeaks:
…tion stack

## Problem
KMS key alias was only checked within CloudFormation stack resources. If the alias was created outside CloudFormation (manually or via another process), the infrastructure code would not discover it and would attempt to create a new key instead of using the existing one.

## Solution
Following DDD and hexagonal architecture principles:

1. Added `describeKmsKey()` method to AWS Provider Adapter
   - Implements port/adapter pattern for KMS queries
   - Enables testing without AWS SDK dependencies

2. Enhanced CloudFormation Discovery to check for KMS aliases
   - Checks `FriggKMSKeyAlias` resources in CloudFormation stack
   - Queries AWS API for expected alias name (based on serviceName/stage)
   - Works even when alias exists outside CloudFormation management

3. Fixed resource extraction to always run AWS queries
   - Changed to call `_extractFromResources()` even with empty resource list
   - Ensures AWS API queries run regardless of CloudFormation stack state

4. Added comprehensive test coverage (TDD)
   - 6 tests covering all KMS alias discovery scenarios
   - Tests verify behavior with mocked dependencies
   - All tests passing

## Impact
- Uses existing KMS keys when alias exists, avoiding duplicate key creation
- Reduces infrastructure costs by reusing existing resources
- Improves idempotency of infrastructure deployment
- Maintains backward compatibility with existing deployments

### DIFF
--- a/packages/devtools/infrastructure/domains/shared/cloudformation-discovery.js
+++ b/packages/devtools/infrastructure/domains/shared/cloudformation-discovery.js
@@ -14,8 +14,10 @@
  */
 
 class CloudFormationDiscovery {
-    constructor(provider) {
+    constructor(provider, config = {}) {
         this.provider = provider;
+        this.serviceName = config.serviceName;
+        this.stage = config.stage;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


seanspeaks:

PR #469, opened 2025-10-23
Fixing the VPC detection bug, it should look for the VPC directly

seanspeaks:


seanspeaks:

PR #472, opened 2025-10-23
fix(cli): propagate stage parameter to resource discovery via SLS_STAGE

seanspeaks:
## Problem
When running `frigg build --stage=qa`, the discovery flow was not using the correct stage. It was defaulting to 'dev' and looking for CloudFormation stack 'quo-integrations-dev' instead of 'quo-integrations-qa'.

## Root Cause
The build command passed --stage to serverless but did not set the SLS_STAGE environment variable that the discovery service reads to determine the stage.

## Solution (TDD/DDD Approach)
Following Test-Driven Development:
1. Added failing tests to document expected behavior
2. Implemented fix in build command to set SLS_STAGE environment variable
3. Added integration test to document CLI-to-discovery contract

Changes:
- build-command/index.js: Set SLS_STAGE in spawned process environment
- build.test.js: Added tests verifying SLS_STAGE is set for qa, production, and dev
- resource-discovery.test.js: Added test documenting SLS_STAGE contract

## DDD/Hexagonal Architecture Alignment
- Maintains separation between CLI (application layer) and discovery (domain layer)
- Uses environment variable as the port between layers
- Discovery service remains cloud-agnostic and testable

### DIFF
--- a/packages/devtools/frigg-cli/__tests__/unit/commands/build.test.js
+++ b/packages/devtools/frigg-cli/__tests__/unit/commands/build.test.js
@@ -157,6 +157,34 @@ describe('CLI Command: build', () => {
       delete process.env.TEST_VAR;
     });
 
+    it('should set SLS_STAGE environment variable to match stage option', async () => {
+      await buildCommand({ stage: 'qa' });
+
+      const call = spawnSync.mock.calls[0];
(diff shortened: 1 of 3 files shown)


seanspeaks:

PR #473, opened 2025-10-23
fix(cli): propagate stage parameter to resource discovery in deploy command

seanspeaks:
## Problem
The deploy command had the same issue as the build command - when running `frigg deploy --stage=qa`, the discovery flow was not using the correct stage. It was defaulting to 'dev' and looking for CloudFormation stack 'example-integrations-dev' instead of 'example-integrations-qa'.

## Root Cause
The deploy command passed --stage to serverless but did not set the SLS_STAGE environment variable that the discovery service reads to determine the stage. The filtered environment was missing SLS_STAGE.

## Solution (TDD/DDD Approach)
Following Test-Driven Development:
1. Created comprehensive test suite for deploy command (was missing)
2. Added failing tests to verify SLS_STAGE propagation
3. Implemented fix to set SLS_STAGE in spawned process environment

Changes:
- deploy-command/index.js: Set SLS_STAGE in spawned process environment
- deploy.test.js: New comprehensive test suite with 20+ test cases
  - Tests for stage propagation (qa, production, dev)
  - Tests for SLS_STAGE environment variable
  - Tests for environment variable filtering
  - Tests for error handling

## DDD/Hexagonal Architecture Alignment
- Maintains separation between CLI (application layer) and discovery (domain layer)
- Uses environment variable as the port between layers
- Consistent with build command implementation
- Discovery service remains cloud-agnostic and testable

### DIFF
--- a/packages/devtools/frigg-cli/__tests__/unit/commands/deploy.test.js
+++ b/packages/devtools/frigg-cli/__tests__/unit/commands/deploy.test.js
@@ -1,0 +1,281 @@
+/**
+ * Test suite for deploy command
+ *
+ * Tests the serverless deployment functionality including:
+ * - Command execution with spawn
+ * - Stage option handling
+ * - Environment variable filtering and propagation
+ * - SLS_STAGE propagation for resource discovery
+ * - Error handling
+ */
+
+// Mock dependencies BEFORE requiring modules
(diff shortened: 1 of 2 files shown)


seanspeaks:

PR #498, opened 2025-11-07
fix(core): Module reference keys, user linking, and credential repository fixes

seanspeaks:
## Summary

This PR contains several important fixes and features for the Frigg core framework:

### 🔑 Module Reference Key Fix
- **fix(core): attach modules using Definition.modules keys via getName()** - Modules are now attached to integrations using the reference key from `Definition.modules` (e.g., `this.quo`) rather than the module's internal name (e.g., `this['quo-attio']`). This allows integrations to use friendly property names while modules can have unique internal identifiers.

### 👥 User Linking & Organization Support
- **feat: Add linkIndividualToOrganization to all user repositories** - New method across MongoDB, PostgreSQL, and DocumentDB repositories to link individual users to organization users
- **fix: Auto-link disconnected users by default with opt-in strict mode** - Automatically links individual users to organization users when both exist
- **fix: Allow organization users to access entities owned by linked individual users** - Uses `User.ownsUserId()` pattern for proper ownership validation

### 🔐 Credential Repository Fixes
- **fix: authIsValid preserved on credential update** - Fixed bug where `authIsValid` was being overwritten with `undefined` when updating credentials (all 3 database repositories)
- **fix: DocumentDB userId converted to ObjectId** - Fixed missing `toObjectId()` conversion for userId in DocumentDB credential operations
- **fix: Support legacy "user" field for backward compatibility** - Credential identifiers now accept both `userId` (preferred) and `user` (legacy)
- **fix: Support both 'user' and 'userId' in module entity identifiers** - Consistent handling across the codebase

### 🏗️ Infrastructure & Schema Updates
- **feat(core): add Process table migration** - New Process model for tracking background processes
- **fix(core): remove duplicate Process table migration and add resolve endpoint**
- **Updated integration-definition.schema.json** - Added `modules`, `hasUserConfig`, `webhooks`, `display`, and `routes` properties to match real-world integration usage

### 🧹 Code Cleanup
- **chore: remove debug console.log statements from production code** - Cleaned up accidentally committed debug statements

## Test Plan

- [x] All 6 `integration-base.module-keys` tests pass (validates module reference key mapping)
- [x] All 28 `integration-router-multi-auth` tests pass
- [x] Schema validation tests pass for new `modules` property
- [x] Verified against real integration (AttioIntegration from quo-frigg)

## Breaking Changes

None - all changes are backward compatible. Legacy `user` field still supported alongside new `userId`.


packages/core/integrations/integration-base.js
```diff
@@ -203,22 +211,68 @@ class IntegrationBase {
...
+                    const definitionName = typeof definition.getName === 'function'
+                        ? definition.getName()
+                        : definition.moduleName;
+                    if (definitionName) {
+                        console.log(`[_appendModules] Mapping: ${definitionName} → ${key}`);
+                        moduleNameToKey[definitionName] = key;
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
seanspeaks: Not actually how the implementation works. Check packages/schema update to understand more

seanspeaks:
:rocket: PR was released in [`v2.0.0-next.57`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.57) :rocket:

packages/core/credential/repositories/credential-repository-documentdb.js
```diff
@@ -143,7 +143,7 @@ class CredentialRepositoryDocumentDB extends CredentialRepositoryInterface {
         }
 
         const plainDocument = {
-            userId: identifiers.userId,
+            userId: toObjectId(identifiers.userId),
```
d-klotz, line 146: I would not change that unless you have tested it with documentDB.

packages/core/credential/repositories/credential-repository-mongo.js
```diff
@@ -121,7 +121,7 @@ class CredentialRepositoryMongo extends CredentialRepositoryInterface {
                 data: {
                     userId: existing.userId,
                     externalId: existing.externalId,
-                    authIsValid: authIsValid,
+                    authIsValid: authIsValid !== undefined ? authIsValid : existing.authIsValid,
```
d-klotz, line 124: Are you sure this is correct? If authIsValid I'd rather say it's false then setting it to what it was in the old DB state (true).

packages/core/credential/repositories/credential-repository-postgres.js
```diff
@@ -111,7 +111,8 @@ class CredentialRepositoryPostgres extends CredentialRepositoryInterface {
         if (!identifiers)
             throw new Error('identifiers required to upsert credential');
 
-        if (!identifiers.userId) {
+        // Support both userId (preferred) and user (legacy) for backward compatibility
+        if (!identifiers.userId && !identifiers.user) {
```
d-klotz, line 115: We should only use one, we don't have legacy yet.
I'd vote for keeping userId.

packages/core/credential/repositories/credential-repository-postgres.js
```diff
@@ -154,7 +155,8 @@ class CredentialRepositoryPostgres extends CredentialRepositoryInterface {
 
         const created = await this.prisma.credential.create({
             data: {
-                userId: this._convertId(identifiers.userId),
+                // Use userId from where clause (supports both userId and user fields)
+                userId: where.userId,
```
d-klotz, line 159: why not using it from the identifiers?

packages/core/database/index.js
```diff
@@ -12,26 +12,53 @@
...
 
 module.exports = {
-    mongoose,
-    IndividualUser,
-    OrganizationUser,
-    UserModel,
-    WebsocketConnection,
-    // Prisma
+    // Lazy-loaded mongoose exports (only load when accessed)
+    get mongoose() {
```
d-klotz, line 31: We don't need that anymore, do we?

packages/core/database/utils/prisma-runner.js
```diff
@@ -373,6 +373,76 @@ async function runPrismaDbPush(verbose = false, nonInteractive = false) {
...
 }
 
+/**
+ * Runs Prisma migrate resolve to mark a migration as applied or rolled back
+ * @param {string} migrationName - Name of the migration to resolve (e.g., '20251112195422_update_user_unique_constraints')
+ * @param {'applied'|'rolled-back'} action - Whether to mark as applied or rolled back
+ * @param {boolean} verbose - Enable verbose output
+ * @returns {Promise<Object>} { success: boolean, output?: string, error?: string }
+ */
+async function runPrismaMigrateResolve(migrationName, action = 'applied', verbose = false) {
```
d-klotz, line 383: Why is this needed? Prisma should do it out of the box.

packages/core/integrations/integration-base.js
```diff
@@ -203,22 +203,53 @@ class IntegrationBase {
...
+        // Build reverse mapping: definition.getName() → referenceKey
+        // e.g., 'quo-attio' → 'quo', 'attio' → 'attio'
+        const moduleNameToKey = {};
+        if (this.constructor.Definition?.modules) {
+            for (const [key, moduleConfig] of Object.entries(this.constructor.Definition.modules)) {
+                const definition = moduleConfig.definition;
+                if (definition) {
+                    // Use getName() if available, fallback to moduleName
+                    const definitionName = typeof definition.getName === 'function'
+                        ? definition.getName()
```
d-klotz, line 228: Why do we have to use `getName` method and not simply the `name` prop?

packages/core/integrations/integration-router.js
```diff
@@ -601,11 +601,10 @@ function setEntityRoutes(router, authenticateUser, useCases) {
     router.route('/api/entities/:entityId/test-auth').get(
         catchAsyncError(async (req, res) => {
             const user = await authenticateUser.execute(req);
-            const userId = user.getId();
             const params = checkRequiredParams(req.params, ['entityId']);
             const testAuthResponse = await testModuleAuth.execute(
                 params.entityId,
-                userId
+                user // Pass User object for proper validation
```
d-klotz, line 607: Why do we need the whole user object?

packages/core/integrations/integration-router.js
```diff
@@ -639,12 +637,11 @@ function setEntityRoutes(router, authenticateUser, useCases) {
     router.route('/api/entities/:entityId/options').post(
         catchAsyncError(async (req, res) => {
             const user = await authenticateUser.execute(req);
-            const userId = user.getId();
             const params = checkRequiredParams(req.params, ['entityId']);
 
             const entityOptions = await getEntityOptionsById.execute(
                 params.entityId,
-                userId
+                user // Pass User object
```
d-klotz, line 644: Passing the whole object makes things more difficult to test.

d-klotz: And most likely we dont need the whole object.

d-klotz: Yeah, just check the tests, where we have to instantiate the user object multiple times.

packages/core/modules/use-cases/get-entity-options-by-id.js
```diff
@@ -12,11 +12,18 @@ class GetEntityOptionsById {
...
-     * @param {string} userId
-     * @param {string} entityId
+     * Retrieve entity options for a given entity
+     *
+     * @param {string|number} entityId - Entity ID to retrieve options for
+     * @param {string|number|import('../../user/user').User} userIdOrUser - User ID or User object for validation
+     * @returns {Promise<Object>} Entity options
      */
-    async execute(entityId, userId) {
+    async execute(entityId, userIdOrUser) {
```
d-klotz, line 21: This is confusing, either we need the user or the userId.

packages/core/modules/use-cases/get-entity-options-by-id.js
```diff
@@ -26,7 +33,12 @@ class GetEntityOptionsById {
             throw new Error(`Entity ${entityId} not found`);
         }
 
-        if (entity.userId !== userId) {
+        // Validate entity ownership
+        const isOwned = typeof userIdOrUser === 'object' && userIdOrUser?.ownsUserId
+            ? userIdOrUser.ownsUserId(entity.userId)
+            : entity.userId?.toString() === userId?.toString();
+
+        if (!isOwned) {
```
d-klotz, line 41: Why do we need this isOwned? this will always throw an error if it's not owned.

packages/core/prisma-postgresql/migrations/20251112195422_update_user_unique_constraints/migration.sql
d-klotz, line 1: We should never change an existing migration. We should rather create a new one with the change.
Also, why delete the Process table creation?

packages/core/credential/repositories/credential-repository-mongo.js
seanspeaks: The intent is to treat this more like a PATCH than a PUT. I.e. If the invoker of upsertCredential omitted the authIsValid value from the params, we shouldn't switch authIsValid to false by default.

packages/core/credential/repositories/credential-repository-documentdb.js
seanspeaks: Mm good point, this was a "hey are we good with the changes in DocumentDB and MongodB" tweak; we can either revert this or test for real (and make sure the unit tests are updated to represent reality)

packages/core/credential/repositories/credential-repository-postgres.js
seanspeaks: We do have legacy, the existing v1 api modules; some map user, some map userId. (This was done because attio uses user vs. userId, or vice versa, in its identifiers. And I wanted to not have to go change each of the api modules in a big sweep pass vs. being a bit more permissive here.

packages/core/credential/repositories/credential-repository-postgres.js
seanspeaks: "where" _does) use it from the identifiers

packages/core/database/index.js
seanspeaks: We don't, good catch, can do a clean sweep PR for any mongoose related things.

packages/core/database/utils/prisma-runner.js
seanspeaks: Prisma does, but the dbMigrate worker only does if we gave it the ability. The worker and api has an abstraction, only running/allowing certain requests to run.

packages/core/integrations/integration-base.js
seanspeaks: At some point in the distant past, someone wanted to use getters and setters for everything I think? So this is the vestige of that. I don't have a strong answer other than "this allows for a dynamic naming convention" if someone wanted to do it.

packages/core/integrations/integration-router.js
seanspeaks: Needs to check for ownership, which you comment on below, so I'll thread there

packages/core/credential/repositories/credential-repository-postgres.js
d-klotz: I'd rather update the api-modules then having to keep both.

packages/core/credential/repositories/credential-repository-postgres.js
d-klotz: yeah but then why change it?

packages/core/credential/repositories/credential-repository-mongo.js
d-klotz: But if authIsValid is false, then even if we patch only a few things, its usage would still not work since the auth is set as invalid.

packages/core/credential/repositories/credential-repository-postgres.js
seanspeaks: Some times it's user sometimes it's userId

packages/core/prisma-postgresql/migrations/20251112195422_update_user_unique_constraints/migration.sql
seanspeaks: This extract CreateTable was breaking the migrations when attempting to run it on a fresh db instance; your hesitation before was well founded and we shouldn't have been creating it. I think the error was actually unrelated to the table existing but rather a percolating up error that was coming across as "table Process does not exist" vs. some issue with the query under the hood. Recently had that with Users and figured out that i was patching the wrong thing.  
  
So I figured these migrations don't work, we should remove them and fix them to work.  


(It's also why I had to bring the prisma resolve method into the dbMigrate worker)

PR #548, opened 2026-03-06
fix(infra): self-heal VPC subnet-route table association drift

d-klotz:
## Summary

This PR fixes a production networking drift issue where Frigg correctly discovered the Lambda route table, but that route table had lost its subnet associations. As a result, Lambda subnets silently fell back to the VPC's main route table instead of using the NAT-backed Lambda route table, which broke outbound connectivity to external APIs.

The fix adds explicit drift detection and self-healing during infrastructure discovery so deployments can repair this state automatically before CloudFormation runs.

## Problem

In the affected AWS topology:

- the Frigg Lambda route table still existed
- the NAT route still existed
- the VPC endpoints still existed
- but the Lambda route table had **0 subnet associations**

When that happens, the subnets use the VPC's main route table instead. For Lambdas running inside the VPC, that means they no longer use the intended NAT egress path and can time out when calling external services.

This is especially tricky because the stack can appear healthy from CloudFormation's perspective even though the real routing state in AWS is broken.

## Root cause

A previous `@friggframework` update changed the logical IDs for `SubnetRouteTableAssociation` resources. CloudFormation did not automatically repair the missing physical associations on later no-op deployments, so the route table remained detached from the Lambda subnets.

### 1. Detect route table association drift

CloudFormation discovery now records the number of subnet associations on the discovered route table via `routeTableAssociationCount`.

This makes the broken state explicit instead of treating the route table as healthy just because it exists.

### 2. Self-heal missing subnet associations before CloudFormation runs

When:

- `vpc.selfHeal` is enabled
- a Lambda route table is discovered
- that route table has `0` subnet associations
- and candidate Lambda subnets are available

Frigg now re-associates the subnets to the correct route table using the EC2 API before CloudFormation continues.

That restores the intended NAT-based routing path automatically during deployment.

### 3. Fix subnet association rebuilding in `VpcBuilder`

`VpcBuilder` was calling `ensureSubnetAssociations()` with `{}` instead of the actual `discoveredResources`, which prevented it from using the real physical subnet IDs in some paths.

This PR fixes that bug so subnet association logic uses the discovered AWS resources correctly.

## Why this matters

This is not a cosmetic infrastructure cleanup. It fixes a real failure mode where:

- the stack still exists
- the route table still exists
- CloudFormation does not recreate anything
- but Lambda outbound networking is broken in production

With this fix, Frigg can now detect and repair that drift automatically instead of leaving Lambdas attached to the wrong route table.

## Impact

This PR makes deployments more resilient by allowing Frigg to recover from route-table/subnet-association drift that CloudFormation alone was not correcting.

In practice, it restores outbound connectivity for VPC Lambdas that depend on NAT routing to reach external APIs.

## Test plan

- deploy with `vpc.selfHeal: true`
- confirm logs show the route table drift being detected
- confirm the missing subnet associations are recreated automatically
- verify the Lambda route table ends up with the expected subnet associations
- verify Lambdas can again reach external APIs through NAT
- verify healthy deployments remain no-op

# Version

Published prerelease version: `v2.0.0-next.74`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  - `@friggframework/devtools`
    - fix(infra): self-heal VPC subnet-route table association drift [#548](https://github.com/friggframework/frigg/pull/548) ([@d-klotz](https://github.com/d-klotz))
    - refactor: replace mongodb dependency by bson [#547](https://github.com/friggframework/frigg/pull/547) ([@d-klotz](https://github.com/d-klotz))
  
  #### Authors: 1
  
</details>

### DIFF
--- a/packages/devtools/infrastructure/domains/networking/vpc-builder.js
+++ b/packages/devtools/infrastructure/domains/networking/vpc-builder.js
@@ -977,7 +977,7 @@ class VpcBuilder extends InfrastructureBuilder {
         }
 
         // Ensure subnet associations
-        this.ensureSubnetAssociations(appDefinition, {}, result);
+        this.ensureSubnetAssociations(appDefinition, discoveredResources, result);
 
         // Create endpoints
         if (endpointsToCreate.includes('s3')) {
--- a/packages/devtools/infrastructure/domains/shared/cloudformation-discovery.js
+++ b/packages/devtools/infrastructure/domains/shared/cloudformation-discovery.js
@@ -562,7 +562,8 @@ class CloudFormationDiscovery {
                             .map(a => a.SubnetId);
                         
                         console.log(`  Route table has ${associatedSubnetIds.length} associated subnets: ${associatedSubnetIds.join(', ')}`);
-                        
+                        discovered.routeTableAssociationCount = associatedSubnetIds.length;
+
                         // Use the associated subnets if available
                         if (associatedSubnetIds.length >= 2) {
                             discovered.privateSubnetId1 = associatedSubnetIds[0];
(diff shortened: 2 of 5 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.74`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.74) :rocket:

PR #572, opened 2026-04-13
fix(core): dedupe integration on re-authorize

d-klotz:
## Summary

- `CreateIntegration` now looks up an existing integration by `(userId, config.type, entityIds)` before creating a new one. If found, the existing record is reused and `testAuth()` runs so stale credentials surface via the usual status/messages pipeline — no `ON_CREATE` re-send, no config overwrite.
- Entity identity already encodes the external account because `ProcessAuthorizationCallback.findOrCreateEntity` dedupes entities by `(userId, moduleName, externalId)`. Re-auth returns the same `entity.id`, so the integration record was the only place duplicates still leaked in.
- Entity comparison is order-insensitive (it's a set).

## Files changed

- `integrations/use-cases/create-integration.js` — dedup logic + `_buildInstance` extraction.
- `integrations/repositories/integration-repository-interface.js` — new abstract `findIntegrationByUserIdTypeAndEntities`.
- `integrations/repositories/integration-repository-{mongo,postgres,documentdb}.js` — implementations. Filter by `userId` + `config.type` at the DB layer, then order-insensitive set equality on entity IDs in JS.
- `integrations/tests/doubles/test-integration-repository.js` — in-memory version.
- `integrations/tests/doubles/dummy-integration-class.js` — adds a `testAuth()` spy (no other tests reference it).
- `integrations/tests/use-cases/create-integration.test.js` — 6 new cases.

## Test plan

- [x] `npm test -- --testPathPattern="create-integration"` — 13/13 pass (6 new).
- [x] `npm test -- --testPathPattern="integrations/tests"` — 129/129 pass.
- [x] Lint: no new errors. Pre-existing abstract-method arg warnings only.

## Notes

- Behaviour change only on the "re-auth" path — fresh creations unchanged.
- The `testAuth()` on reuse writes to `messages.errors` and flips `status` to `ERROR` on failure via `UpdateIntegrationStatus` / `UpdateIntegrationMessages`, so the API response naturally reflects credential validity.

### DIFF
--- a/packages/core/integrations/repositories/integration-repository-documentdb.js
+++ b/packages/core/integrations/repositories/integration-repository-documentdb.js
@@ -148,6 +148,30 @@ class IntegrationRepositoryDocumentDB extends IntegrationRepositoryInterface {
         return doc ? this._mapIntegration(doc) : null;
     }
 
+    async findIntegrationByUserIdTypeAndEntities(userId, type, entityIds) {
+        const userObjectId = toObjectId(userId);
+        if (!userObjectId) return null;
+
+        const targetIds = toObjectIdArray(entityIds || [])
+            .map((value) => fromObjectId(value))
+            .sort();
+
+        const candidates = await findMany(this.prisma, 'Integration', {
+            userId: userObjectId,
+            'config.type': type,
+        });
+
+        const match = candidates.find((doc) => {
+            const current = (doc?.entityIds || [])
+                .map((value) => fromObjectId(value))
+                .sort();
+            if (current.length !== targetIds.length) return false;
+            return current.every((id, idx) => id === targetIds[idx]);
+        });
+
+        return match ? this._mapIntegration(match) : null;
+    }
+
     async updateIntegrationConfig(integrationId, config) {
         if (config === null || config === undefined) {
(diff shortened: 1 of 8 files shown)


d-klotz:
### Code review

Found 1 issue:

1. On the reuse path, `await instance.testAuth()` persists status/messages to the DB via `updateIntegrationStatus.execute(...)` / `updateIntegrationMessages.execute(...)` but does not mutate `this.status` / `this.messages` on the in-memory instance. The next line returns `mapIntegrationClassToIntegrationDTO(instance)`, which reads directly from the instance — so when auth has just failed, the DB row is `ERROR` but the HTTP response still shows the pre-`testAuth` status (typically `ENABLED`) with no errors. This contradicts this PR's own JSDoc promising stale credentials "surface as an `ERROR` status / message".

https://github.com/friggframework/frigg/blob/dd85498e4a9714789ed9e866a4ff717aaadb707e/packages/core/integrations/use-cases/create-integration.js#L49-L57

Fix options: re-read the record from the repository after `testAuth()` and map that, or have `testAuth()` also mutate the instance's `status` / `messages` when it updates the DB.

PR #575, opened 2026-04-15
fix(core): narrow OAuth2Requester isRefreshable when refresh is impossible

d-klotz:
## Summary

For OAuth providers that never issue refresh tokens (canonical case: **Attio** — spec is auth-code-only, `app.attio.com/oauth/token` explicitly rejects `grant_type=refresh_token` with `HTTP 400 "grant_type must be one of [authorization_code, code]"`), Frigg's 401 retry path today makes a **guaranteed HTTP 400** on every retry. In production this produces the primary error-log volume on the Attio queue worker.

The `isRefreshable` flag is already the gate for the 401 retry path at `requester.js:74`:

```javascript
if (!this.isRefreshable || this.refreshCount > 0) {
    await this.notify(this.DLGT_INVALID_AUTH);  // <-- short-circuit when false
} else {
    this.refreshCount++;
    const refreshSucceeded = await this.refreshAuth();
    ...
}
```

The bug is that `isRefreshable` is unconditionally defaulted to `true` at `OAuth2Requester:101`, ignoring whether a `refresh_token` was ever issued. For any provider that never gives one back, the default is wrong.

This PR makes `isRefreshable` honest: **`true` means "I can refresh," `false` means "I can't — go straight to invalidation."** Derivation is consistent at every point where `refresh_token` state is set.

### `packages/core/modules/requester/oauth-2.js`

- **Constructor (Edit 1)** — narrow `isRefreshable` to `false` when `grant_type !== 'client_credentials'` AND `!refresh_token`. Subclass opt-outs via `this.isRefreshable = false` after `super(params)` continue to work (narrowing runs before the subclass body).
- **`setTokens` (Edit 2)** — re-derive `isRefreshable` from current state after any token-response update. Correctly **widens** from false → true on first-time auth for refresh-supporting providers (Pipedrive, Zoho) whose initial auth-code response carries a refresh_token. Idempotent on subsequent refreshes.
- **`refreshAuth` (Edit 3)** — defense-in-depth guard. When `isRefreshable` is false, short-circuit to `notify(DLGT_INVALID_AUTH)` without invoking the token endpoint. Covers direct callers bypassing `requester.js:74`.

### `packages/core/modules/requester/oauth-2.test.js`

- **5 tests** for constructor narrowing: `authorization_code` + no RT → false; `authorization_code` + RT → true; `password` + no RT → false; `client_credentials` regression → true; subclass-override-wins regression.
- **4 tests** for `setTokens` re-derivation: widen (first-time auth), preserve-false (Attio), preserve-true (normal refresh), `client_credentials` passthrough.
- **1 test** for `refreshAuth` guard: `isRefreshable=false` short-circuits and notifies.
- **1 integration test** through `_get`: OAuth2Requester with no `refresh_token` receives a 401 — verifies `refreshAuth`/`refreshAccessToken` are never called, `fetch` invoked exactly once, `DLGT_INVALID_AUTH` fired once.
- **1 pre-existing test updated** (`'should set isRefreshable to true'`) to explicitly supply a `refresh_token`, since the meaning of "default" changed.

## Behavior matrix

| Scenario | Constructor | First `setTokens` | On 401 |
|---|---|---|---|
| Attio (no RT ever) | `false` | stays `false` | notify `DLGT_INVALID_AUTH`, no POST |
| Pipedrive (RT on first auth) | `false` | widens to `true` | refresh succeeds |
| Zoho (RT on first auth) | `false` | widens to `true` | refresh succeeds |
| `client_credentials` | `true` | stays `true` | refresh via `getTokenFromClientCredentials` |
| Subclass `isRefreshable=false` + any RT | `false` (subclass wins) | stays `false` (widen can override only when constructor defaults) | notify `DLGT_INVALID_AUTH` |

## Test plan

- [x] 38 unit tests pass (28 pre-existing + 10 new + 1 updated + 1 integration)
- [x] Full core suite: zero new failures (the 13 pre-existing `map-integration-dto` failures are unrelated, untouched)
- [ ] E2E verification on a live Attio dead-token scenario: trigger 401 → confirm NO `[Frigg] Starting token refresh` / `Token refresh failed` / `POST …/oauth/token 400` in logs, confirm `Credential.authIsValid = false` in DB

## Context

This is **Gap A of three** framework-level fixes from an [RCA](https://github.com/friggframework/frigg) on Attio dead-token loops.

- **Gap C** (restore `Integration.status` on successful re-auth) landed in friggframework/frigg#574.
- **Gap B** (auto-flip `Integration.status` to `DISABLED` on `DLGT_INVALID_AUTH`) is a separate upcoming PR.

Each PR has zero file overlap with the others and merges independently.

### DIFF
--- a/packages/core/integrations/integration-router.js
+++ b/packages/core/integrations/integration-router.js
@@ -190,6 +190,7 @@ function createIntegrationRouter() {
     const processAuthorizationCallback = new ProcessAuthorizationCallback({
         moduleRepository,
         credentialRepository,
+        integrationRepository,
         moduleDefinitions:
             getModulesDefinitionFromIntegrationClasses(integrationClasses),
     });
--- a/packages/core/integrations/repositories/integration-repository-documentdb.js
+++ b/packages/core/integrations/repositories/integration-repository-documentdb.js
@@ -26,6 +26,15 @@ class IntegrationRepositoryDocumentDB extends IntegrationRepositoryInterface {
         return records.map((doc) => this._mapIntegration(doc));
     }
 
+    async findIntegrationsByEntityId(entityId) {
+        const objectId = toObjectId(entityId);
+        if (!objectId) return [];
+        const records = await findMany(this.prisma, 'Integration', {
+            entityIds: objectId,
+        });
+        return records.map((doc) => this._mapIntegration(doc));
+    }
+
     async deleteIntegrationById(integrationId) {
         const objectId = toObjectId(integrationId);
         if (!objectId) return { acknowledged: true, deletedCount: 0 };
--- a/packages/core/integrations/repositories/integration-repository-interface.js
+++ b/packages/core/integrations/repositories/integration-repository-interface.js
@@ -122,6 +122,23 @@ class IntegrationRepositoryInterface {
     async updateIntegrationConfig(integrationId, config) {
         throw new Error('Method updateIntegrationConfig must be implemented by subclass');
     }
+
+    /**
+     * Find all integrations whose entity set includes the given entity ID.
+     *
+     * Used by the authorization callback flow to walk up from a re-authorized
+     * entity to its parent integrations so that any in a broken state (ERROR,
+     * DISABLED) can be restored to ENABLED.
+     *
+     * @param {string|number} entityId - Entity ID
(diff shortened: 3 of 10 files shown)


d-klotz:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #582, opened 2026-05-05
fix(core): explicitly persist tokens on OAuth2 re-auth

d-klotz:
## Summary

`ProcessAuthorizationCallback` relied solely on the `DLGT_TOKEN_UPDATE` notification chain (`setTokens` → `notify` → `Module.onTokenUpdate` → `upsertCredential`) to persist freshly-issued OAuth tokens. In production this chain has been observed to silently no-op: the OAuth provider issues new tokens (and invalidates the prior `refresh_token`), but the new tokens never land in the DB. The user-visible `/api/authorize` call appears to succeed, the existing credential is reused with stale token data, and downstream API calls fail with 401 once the provider rotates the old refresh_token.

This was diagnosed in a Pipedrive integration: re-installs created new Integration rows pointing at a 3-month-old credential whose tokens had been retired by Pipedrive during today's OAuth flow, but never replaced in our DB. The auth Lambda logs were silent because none of the existing code logs at the relevant boundaries.

## Changes

- **Bootstrap the Module with the existing entity** — `process-authorization-callback.js` now looks up the existing entity via `findEntitiesByUserIdAndModuleName` and passes it to `new Module(...)` so the API requester is preloaded with prior tokens. Replaces the long-standing `// todo: check if we need to pass entity to Module, right now it's null` TODO.
- **Belt-and-suspenders explicit persistence on the OAuth2 path** — explicit `await this.onTokenUpdate(module, moduleDefinition, userId)` after `getToken`. The notification chain stays in place; this ensures persistence happens even when the chain silently no-ops. Idempotent against the notification path because `upsertCredential` is keyed by `externalId`.
- **Diagnostic logging** at entry, after `getToken`, after persistence, and after restore — so future re-auth bugs aren't invisible in CloudWatch.
- **Wrap `POST /api/authorize`** with try/catch + `console.error` so handler failures surface instead of returning silently.
- **`restoreIntegrationsForEntity` returns the count** of flipped integrations for logging.

## Test plan

- [x] Existing 5 `ProcessAuthorizationCallback` re-auth status restoration tests pass
- [x] New test: `persists credentials explicitly on OAuth2 re-auth (belt-and-suspenders)` — locks in that `upsertCredential` is called exactly once on the OAuth2 path
- [ ] Smoke-test in dev: re-install a Pipedrive integration; confirm CloudWatch shows the new `[Frigg] processAuthorizationCallback ...` log lines and that the credential's `updatedAt` matches the OAuth callback time (not a later auth-flip)
- [ ] Verify a re-auth on a credential currently in `ERROR` flips it back to `ENABLED` and the integration resumes syncing

# Version

Published prerelease version: `v2.0.0-next.82`

<details>
  <summary>Changelog</summary>

  #### 🚀 Enhancement
  
    - fix(core): add per-request timeout to Requester to catch silent fetch hangs [#579](https://github.com/friggframework/frigg/pull/579) ([@d-klotz](https://github.com/d-klotz))
  
  #### 🐛 Bug Fix
  
  - `@friggframework/core`
    - fix(core): explicitly persist tokens on OAuth2 re-auth [#582](https://github.com/friggframework/frigg/pull/582) ([@d-klotz](https://github.com/d-klotz))
  
  #### Authors: 1
  
</details>

### DIFF
--- a/packages/core/integrations/integration-router.js
+++ b/packages/core/integrations/integration-router.js
@@ -507,7 +507,8 @@ function setEntityRoutes(router, authenticateUser, useCases) {
             const params = checkRequiredParams(req.query, ['entityType']);
             const module = await getModuleInstanceFromType.execute(
                 userId,
-                params.entityType
+                params.entityType,
+                { state: req.query.state }
             );
             const areRequirementsValid =
                 module.validateAuthorizationRequirements();
--- a/packages/core/integrations/integration-router.js
+++ b/packages/core/integrations/integration-router.js
@@ -507,8 +507,7 @@ function setEntityRoutes(router, authenticateUser, useCases) {
             const params = checkRequiredParams(req.query, ['entityType']);
             const module = await getModuleInstanceFromType.execute(
                 userId,
-                params.entityType,
-                { state: req.query.state }
+                params.entityType
             );
             const areRequirementsValid =
                 module.validateAuthorizationRequirements();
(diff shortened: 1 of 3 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.82`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.82) :rocket:

PR #583, opened 2026-05-05
fix(core): repoint entity credentialId when re-auth produces a different credential

d-klotz:
## Summary

Follow-up to #582. When a user re-authenticates against a different workspace/account of the same provider (e.g. switching between two Pipedrive workspaces under the same Frigg user), the re-auth flow leaves the entity linked to the *prior* workspace's credential while the freshly-issued tokens land on a different credential row.

## Root cause

`upsertCredential` matches/creates a credential keyed by externalId from `getCredentialDetails` (e.g. Pipedrive `userProfile.data.id`). `findOrCreateEntity` matches the entity by its own externalId from `getEntityDetails` (e.g. Pipedrive `userProfile.data.company_id`). When the user has multiple workspaces, both methods can hit different existing rows — `upsertCredential` resolves to credential B's row, `findOrCreateEntity` returns the entity already pointing at credential A.

The OAuth callback then reports `{ credentialId: B, entityId: matched-entity }`, but the entity's `credentialId` foreign key still references A. Downstream integration creation reads the stale link, and the just-issued tokens sit on credential B with no entity pointing to it.

## Observed in dev

- Pipedrive integration 136 (user 16) created via re-auth into workspace `24952048` (cred 188).
- Entity 159 was previously linked to cred 166 (workspace `23384438` from a prior install).
- After OAuth: cred 188 had fresh tokens, cred 166 was unchanged at the OAuth time. Entity 159 still pointed at cred 166. Webhook setup later succeeded only because the runtime auto-refresh on cred 166's *separate* refresh_token happened to still be alive.

## Fix

`findOrCreateEntity` now repoints `existingEntity.credentialId` to the just-upserted `credentialId` via `moduleRepository.updateEntity` whenever they differ. No-ops when they match.

## Tests

- `repoints existing entity credentialId when re-auth produces a different credential` — locks in the new behavior
- `does not repoint when the existing entity already points at the upserted credential` — guards against unnecessary writes
- All 6 existing tests still pass

### DIFF
--- a/packages/core/modules/use-cases/__tests__/process-authorization-callback.test.js
+++ b/packages/core/modules/use-cases/__tests__/process-authorization-callback.test.js
@@ -36,7 +36,7 @@ describe('ProcessAuthorizationCallback — re-auth status restoration', () => {
                 userId: 'user-1',
                 moduleName: 'testmodule',
                 externalId: 'ext-1',
-                credential: 'cred-1',
+                credential: { id: 'cred-1' },
             }),
             findEntitiesByUserIdAndModuleName: jest.fn().mockResolvedValue([
                 {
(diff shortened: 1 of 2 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.83`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.83) :rocket:

PR #588, opened 2026-05-21
feat(core): increment totalSkipped counter on UpdateProcessMetrics

d-klotz:
## Summary

- `UpdateProcessMetrics.execute` accepts a new `skipped` field on the metrics-update object and routes it into the atomic phase as `results.aggregateData.totalSkipped`.
- Mirrors how `success` and `errors` are handled: omitted from the increment map when zero, and counts as atomic work so a skipped-only batch doesn't fall through to `findById`.
- WebSocket broadcasts additionally expose `skippedCount` in the progress payload.

## Why

The UI identity `processed = totalSynced + totalFailed + totalSkipped` was breaking for integrations that legitimately skip records (hash-match, loop protection, dedupe). The old code only had three increment paths — `processed`, `success`, `errors` — so any `skipped: N` passed through `recordBatchMetrics` was silently dropped on the atomic path. The best it could do was leave a `kind: 'skipped_count'` breadcrumb in the bounded `errorDetails` array (capped at 100), which then showed up as `errors` in any UI that surfaced the array.

Concrete case that surfaced this: a HubSpot ↔ Clockwork integration was hash-skipping ~574 records per initial sync (records that originated in the destination system, so the c2h hash matched and the upsert was correctly skipped). The status panel showed:

```
Processed:  3086
Synced:     2512
Failed:        0
Skipped:       0   ← actually 574, dropped by the atomic path
```

After this change the same scenario would write `totalSkipped = 574` and the identity holds.

## Compatibility

- **Pure addition.** Existing callers that don't pass `skipped` see no behaviour change — the existing 17 tests still pass unchanged.
- **WebSocket broadcast** now includes `skippedCount` alongside `successCount`/`errorCount`. Additive; no removed fields.
- No DB schema change. `totalSkipped` is created on first increment by `jsonb_set` / `\$inc` `create_missing=true`.

## Tests

3 new tests in `update-process-metrics.test.js`:

- `routes skipped to applyProcessUpdate.increment as totalSkipped`
- `omits totalSkipped from the increment map when skipped is zero`
- `treats a skipped-only batch as atomic work (does not short-circuit)`

Full suite for `update-process-metrics.test.js`: **20/20 green** (17 prior + 3 new).

Wider `packages/core` baseline: 64 failures pre-existing on `next` at `75dc5e65`. Same 64 failures with my change applied — confirmed via `git stash` round-trip. None of them touch this code path.

## Test plan

- [x] New unit tests cover the additive code paths.
- [x] Confirmed baseline failure count unchanged.
- [ ] Reviewer: confirm WebSocket subscribers don't reject unknown additional fields in the progress payload.

# Version

Published prerelease version: `v2.0.0-next.86`

<details>
  <summary>Changelog</summary>

  #### 🚀 Enhancement
  
  - `@friggframework/core`
    - feat(core): increment totalSkipped counter on UpdateProcessMetrics [#588](https://github.com/friggframework/frigg/pull/588) ([@d-klotz](https://github.com/d-klotz))
  
  #### 🐛 Bug Fix
  
    - fix(core): respect refresh outcome before marking integration ERROR on 401 [#587](https://github.com/friggframework/frigg/pull/587) ([@d-klotz](https://github.com/d-klotz))
  
  #### Authors: 1
  
</details>

### DIFF
--- a/packages/core/integrations/use-cases/update-process-metrics.js
+++ b/packages/core/integrations/use-cases/update-process-metrics.js
@@ -28,8 +28,9 @@
  * const updateMetrics = new UpdateProcessMetrics({ processRepository, websocketService });
  * await updateMetrics.execute(processId, {
  *   processed: 100,
- *   success: 95,
+ *   success: 92,
  *   errors: 5,
+ *   skipped: 3,
  *   errorDetails: [{ contactId: 'abc', error: 'Missing email', timestamp: '...' }]
  * });
  */
@@ -54,6 +55,11 @@ class UpdateProcessMetrics {
      * @param {number} [metricsUpdate.processed=0] - Records processed in this batch
      * @param {number} [metricsUpdate.success=0] - Successful records
      * @param {number} [metricsUpdate.errors=0] - Failed records
+     * @param {number} [metricsUpdate.skipped=0] - Intentionally-skipped
+     *     records (hash-match, dedupe, loop protection, etc.). Increments
+     *     `results.aggregateData.totalSkipped`. Distinct from errors so the
+     *     UI can show `processed = synced + failed + skipped` without
+     *     conflating intentional skips with failures.
      * @param {Array} [metricsUpdate.errorDetails=[]] - Error details array
      * @returns {Promise<Object>} Updated process record
      * @throws {Error} If process not found or update fails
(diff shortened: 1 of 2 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.100`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.100) :rocket:

PR #589, opened 2026-05-22
docs(architecture): ADR for plugins and extensions taxonomy

seanspeaks:
## Summary

Adds `docs/architecture/ADR-EXTENSIONS.md` — a proposed three-tier taxonomy for what's pluggable in Frigg:

- **Core Plugins** — required infrastructure with defaults (provider, queue, encryption, DB). Already covered by [ADR-MULTI-PROVIDER-SUPPORT](https://github.com/friggframework/frigg/blob/next/docs/architecture/ADR-MULTI-PROVIDER-SUPPORT.md); this ADR names the category for contrast.
- **Application Extensions** — optional, declarative additions to the core app, listed on `appDefinition.extensions`. Grounded in the `packages/core/extensions/` prototype on [`claude/frigg-netlify-exploration-aY2Bh`](https://github.com/friggframework/frigg/tree/claude/frigg-netlify-exploration-aY2Bh/packages/core/extensions) and the reference [`@friggframework/extension-db-credentials`](https://github.com/friggframework/frigg/tree/claude/frigg-netlify-exploration-aY2Bh/packages/extensions/db-credentials) package.
- **Integration Extensions** — off-the-shelf handler bundles (webhooks, sync routines) shipped with API modules, bound into an integration's `Definition.extensions`. Grounded in the `this.extensions` prototypes in [`lefthookhq/frigg-2.0-prototyping`](https://github.com/lefthookhq/frigg-2.0-prototyping/blob/main/backend/src/integrations/HubSpotIntegration.js), `stack-global--frigg-2.0`, and `vartopia--frigg-2.0`.

## What the ADR resolves

- **Vocabulary collision.** "Plugin" and "extension" are currently used for at least three different things across the codebase and prototypes.
- **Shape mismatch.** The Tier 2 loader expects `extensions: [...]` (array); the Tier 3 prototypes use `extensions: { name: {...} }` (object). The ADR keeps both — they model different problems — and explains why.
- **`this.extensions` vs `static Definition.extensions`.** Recommends moving Tier 3 wiring out of the constructor and onto the static Definition, with string-name handler binding done by the framework at instantiation.

## Status

- **Status: Proposed** — open for review and revision before acceptance.
- No code changes. Pure design doc.
- Implementation phases sketched at the end of the ADR (Tier 2 land → Tier 3 formalize → docs → deprecate old shape).

## Open questions in the ADR

1. Handler-binding-by-string trade-offs vs. a factory form
[...]

Issue #89, opened 2023-01-10
strKey is not defined


joncodo:
ReferenceError: strKey is not defined
    at get (/Users/jon/Code/lefthook/canva--frigg/backend/.webpack/service/src/handlers/http/webpack:/frigg-app-back-end/node_modules/@friggframework/assertions/get.js:17:1)

asertions/get.js

```
const get = (o, key, defaultValue) => {
    const value = lodashGet(o, key, defaultValue);

    if (value !== undefined) {
        return value;
    }

    if (defaultValue === undefined) {
        throw new RequiredPropertyError({
            parent: this,
            key: strKey,
        });
    }

    return defaultValue;
};
```

strKey is not declared in this method. You only see this error on run so maybe I am the first one to hit it.

seanspeaks:
Resolved

PR #590, opened 2026-05-22
feat(core): Tier 3 Integration Extensions — Definition.extensions

d-klotz:
## What this is

Right now, every integration that wants to receive webhooks rewrites the same plumbing: an HTTP receiver, signature verification, a portal-ID lookup to find the right per-account integration record, and a hand-off to a queue worker. Copy-paste, fifty lines at a time, integration after integration.

This PR adds the framework half of Sean's [Plugins & Extensions ADR](https://github.com/friggframework/frigg/pull/589) so an API module can ship that plumbing **once** as a reusable bundle, and an integration consumes it declaratively:

```js
class HubSpotIntegration extends IntegrationBase {
    static Definition = {
        name: 'hubspot',
        modules: { hubspot: { definition: hubspot.Definition } },
        extensions: {
            hubspotWebhooks: {
                extension: hubspot.extensions.webhooks,
                handlers: { HUBSPOT_WEBHOOK: 'onHubSpotEvent' },
            },
        },
    };

    async onHubSpotEvent({ data }) {
        // pure business logic — signature, lookup, queueing all done by the extension
    }
}
```

That's it. Routes are mounted automatically, the receiver verifies signatures, queue dispatch is wired, and the integration developer only writes the reaction. The same shape works for HubSpot CRM cards, Timeline events, Slack events, Notion webhooks, whatever — once the API module ships the bundle.

## Why now

Daniel opened [api-module-library#89](https://github.com/friggframework/api-module-library/pull/89) to add three HubSpot subscription-management methods to unblock a downstream Clockwork integration. Sean pushed back: those methods are app-level (developer API key), not per-portal (OAuth) — they don't fit the per-portal `Api` class contract, and the *real* missing piece is the extension layer that bundles the webhook receiver/dispatcher in the first place. He'd been [noodling on this concept since 2023](https://lefthookhq.slack.com/archives/C017RNCQV70/p1699476664994709) and wrote the [ADR](https://github.com/friggframework/frigg/blob/adr/extensions/docs/architecture/ADR-EXTENSIONS.md) to formalize it.

This PR lands the framework piece. The HubSpot bundle that consumes it (and the place where #89's three primitives belong, as internal scaffolding) is a follow-up in the api-module-library repo.

## How it resolves the app-level vs account-level question

The objection on #89 was: webhook subscription management is app-level, but the api-module is per-portal. The extension layer makes the right split:

| Concern | Where it lives | Identity used |
|---|---|---|
| Subscription registration (app-wide) | Extension's optional build-time helper | Dev API key (app-level) |
| Webhook reception | Extension-declared route, mounted by framework | None (HMAC verified inside extension) |
| Per-portal dispatch | Framework's `findIntegrationByPortalId` lookup | Per-portal OAuth token loaded for the bound handler |

#89's three methods (list/create/delete subscriptions) become the **internal primitives** the build helper calls, not public api-module surface that integration developers touch.

### The contract — `packages/core/integrations/extension.js` (new)

An extension bundle is a plain object:

```js
module.exports = {
    name: 'hubspot-webhooks',
    routes: [{ path: '/webhooks', method: 'POST', event: 'HUBSPOT_WEBHOOK_RECEIVED' }],
    events: {
        HUBSPOT_WEBHOOK_RECEIVED: { handler: async function ({ req, res }) { /* verify + queue */ } },
        HUBSPOT_WEBHOOK: { handler: async function ({ data }) { /* default no-op */ } },
    },
};
```

Three helpers ship with it: `validateExtensionBinding`, `getExtensionRoutes`, `getExtensionWorkers`. All exported from `@friggframework/core`.

### The wiring — three small seams

1. **`IntegrationBase._mergeExtensions()`** (called inside `initialize()`) walks `static Definition.extensions`, validates each binding, and merges extension events into `this.events` with handlers bound to the live instance. Resolution priority:
   - Subclass-defined `this.events[event]` (constructor) wins
   - Else `binding.handlers[event]` → method name string, resolved on `this`
   - Else extension's own default handler
   - Else throw at boot with clear context

2. **`integration-defined-routers.js`** mounts extension routes alongside `Definition.routes` via the existing `loadRouterFromObject` — no admin auth (extensions own their own signature verification).

3. **`loadRouterFromObject` and the worker's dry-instance branch** now call `await integrationInstance.initialize()` before dispatch so extension events are merged into `this.events` before `IntegrationEventDispatcher.findEventHandler` runs. The dispatcher itself is untouched.

### Fail-loud defaults

The new code refuses to silently misroute:

- **Two extensions claiming the same event name** → throw at `initialize()` with both binding names
- **Two routes claiming the same `method + path`** → throw at boot with both sources (Definition.routes / extension X / extension Y)
- **Binding handler references a non-existent method** → throw with integration + binding + method name
- **`binding.handlers` typos** (referencing an event the extension doesn't declare) → throw with "check for typos"
- **`findIntegrationByPortalId` ambiguous resolution** (one externalId, multiple integrations) → throw rather than blindly pick `[0]` — would otherwise be a cross-tenant data leak
- **Subclass shadowing a binding-declared handler** → `console.warn` so the author knows their override is ignored (subclass wins by design)

### The reverse-lookup helper

`IntegrationBase.findIntegrationByPortalId(externalId, moduleName?)` composes `findEntity` + `findIntegrationsByEntityId` so an extension's default webhook handler can resolve an inbound `portalId` (Slack `team_id`, etc.) to a per-account integration ID without each api-module reimplementing the lookup.

## What this looks like end-to-end

Boot logs print the extension routes alongside Definition routes:

```
│ Configuring routes for hubspot Integration:
│ POST /api/hubspot-integration/some-existing-route
│ POST /api/hubspot-integration/webhooks  (extension: hubspot-webhooks)
│
```

A webhook arrives → extension's signature verification fires → portalId lookup → queued → worker hydrates the right integration record → bound `onHubSpotEvent` method on the integration instance executes with the correct per-portal OAuth credentials in context. The integration developer wrote ~10 lines of business logic.

## What's NOT in this PR (scope)

- **The HubSpot bundle** (`hubspot.extensions.webhooks`, `crmCards`, `timeline`) lives in `friggframework/api-module-library` — separate repo, follow-up PR.
- **No migration of existing integrations** to the new pattern. `Definition.webhooks: true` and `onWebhookReceived`/`onWebhook` continue to work unchanged. The two patterns coexist; see [EXTENSIONS.md](packages/core/integrations/EXTENSIONS.md) for when to use which.
- **No Tier 2 Application Extensions** (`appDefinition.extensions: [...]`) — that's the [`claude/frigg-netlify-exploration-aY2Bh`](https://github.com/friggframework/frigg/tree/claude/frigg-netlify-exploration-aY2Bh/packages/core/extensions) branch's work and a separate PR.

## Known limitations / follow-ups

These are intentionally deferred to keep this PR focused — none change behavior, only optimize or tighten:

1. **Per-class merged-event cache.** Today `_mergeExtensions` + `loadDynamicUserActions` re-run on every HTTP request (via `loadRouterFromObject → initialize`). The merge result is fully derivable from static Definition; cache it per class at first call. Matters for webhook firehoses.
2. **`addError` on ephemeral `loadDynamicUserActions` failures.** Pre-existing; calling `initialize()` per request makes the in-memory `status = 'ERROR'` side-effect more visible. Move dynamic-action loading out of the hot path or stop calling `addError` for transient failures.
3. **`route.middleware: []` on extension routes.** Extensions own their auth, but there's no declarative seam for middleware — an author who forgets signature verification leaves an open endpoint. Cheap follow-up.
4. **`eventDef.type` whitelist** so an extension can't accidentally declare `type: 'USER_ACTION'` and have its webhook events leak into `loadUserActions()` UI output.
5. **`extension.name` uniqueness across bindings.** Currently two bindings of the same extension produce identical boot-log entries (`(extension: foo)` × 2), making debugging harder. Cosmetic.
6. **Stdout-capture tests** to lock the boot-log format string so future refactors don't silently break log consumers.
7. **Repository access via use case, not lazy require.** `findIntegrationByPortalId` does a `require()` inside the method (matches the established `queueWebhook` pattern in the same file), but eventually we should route this through a use case per the project's hexagonal-architecture rules.

## Test plan

- [x] 69 new unit tests covering the contract, validation, merge priority, conflict detection, route mounting, dispatch path, and helper semantics. All pass.
- [x] No regressions in `packages/core/integrations/` or `packages/core/handlers/`. Two pre-existing failing suites (`webhook-flow.integration.test.js` and `integration-mapping-repository-documentdb-encryption.test.js`) were verified to fail identically on clean `next` — unrelated AWS SDK + Jest ESM issues.
- [ ] End-to-end smoke against a real HubSpot dev account — deferred to the api-module-library bundle PR (this PR doesn't have an extension to point at yet).

## References

- ADR: [docs/architecture/ADR-EXTENSIONS.md](https://github.com/friggframework/frigg/blob/adr/extensions/docs/architecture/ADR-EXTENSIONS.md) (in [#589](https://github.com/friggframework/frigg/pull/589))
- Triggering PR: [api-module-library#89](https://github.com/friggframework/api-module-library/pull/89) — HubSpot webhook subscriptions
- 2023 Loom + Slack thread where Sean originally sketched the concept: linked in #589
- Quickstart: [packages/core/integrations/EXTENSIONS.md](packages/core/integrations/EXTENSIONS.md)

<details>
  <summary>📦 Published PR as canary version: <code>2.0.0--canary.590.c1af198.0</code></summary>
  <br />

  ```bash
  npm install @friggframework/core@2.0.0--canary.590.c1af198.0
  npm install @friggframework/devtools@2.0.0--canary.590.c1af198.0
  npm install @friggframework/eslint-config@2.0.0--canary.590.c1af198.0
  npm install @friggframework/prettier-config@2.0.0--canary.590.c1af198.0
  npm install @friggframework/schemas@2.0.0--canary.590.c1af198.0
  npm install @friggframework/serverless-plugin@2.0.0--canary.590.c1af198.0
  npm install @friggframework/test@2.0.0--canary.590.c1af198.0
  npm install @friggframework/ui@2.0.0--canary.590.c1af198.0
  # or 
  yarn add @friggframework/core@2.0.0--canary.590.c1af198.0
  yarn add @friggframework/devtools@2.0.0--canary.590.c1af198.0
  yarn add @friggframework/eslint-config@2.0.0--canary.590.c1af198.0
  yarn add @friggframework/prettier-config@2.0.0--canary.590.c1af198.0
  yarn add @friggframework/schemas@2.0.0--canary.590.c1af198.0
  yarn add @friggframework/serverless-plugin@2.0.0--canary.590.c1af198.0
  yarn add @friggframework/test@2.0.0--canary.590.c1af198.0
  yarn add @friggframework/ui@2.0.0--canary.590.c1af198.0
  ```
</details>

# Version

Published prerelease version: `v2.0.0-next.88`

<details>
  <summary>Changelog</summary>

  #### 🚀 Enhancement
  
  - `@friggframework/core`, `@friggframework/devtools`
    - feat(core): Tier 3 Integration Extensions — Definition.extensions [#590](https://github.com/friggframework/frigg/pull/590) ([@d-klotz](https://github.com/d-klotz))
    - feat(core): increment totalSkipped counter on UpdateProcessMetrics [#588](https://github.com/friggframework/frigg/pull/588) ([@d-klotz](https://github.com/d-klotz))
  
  #### 🐛 Bug Fix
  
    - fix: query by entity binding instead of user in FindIntegrationContextByExternalEntityId (FRI-498) [#593](https://github.com/friggframework/frigg/pull/593) ([@claude](https://github.com/claude) [@d-klotz](https://github.com/d-klotz))
  
  #### Authors: 2
  
  - Claude ([@claude](https://github.com/claude))
</details>


seanspeaks:
The findInregrarionByPortalId doesn't belong in core, that belongs in HubSpot as part of the extension. "Portal id" is specific to HubSpot 

Core COULD export a helper that is findIntegrationByEntityExternalId (or "list" etc)

Then HubSpot would use that and pass in the portal id. 

Might be a better fit. 

Needs to be used bt a healthy number of integrations who only send their account as the identifier. To wit: Google Workspace, Asana, Slack, Microsoft Teams, HubSpot just to name a few. 

Anywho, that comment, plus:

https://github.com/friggframework/frigg/pull/591

Plus I think it's important to call out that there are other primitives and events we'd want to surface in a frigg integration definition and therefore in an api module extension. 
1. Queues. Custom queues (and corresponding handlers) as needed. 
2. Crons. Event Bridge or elsewhere
3. User Actions. 
4. And I think more just blanking at the moment

seanspeaks:
A note on `findIntegrationByPortalId` placement and naming — not blocking, but worth resolving before this merges since it sets a precedent for every API module that ships a webhook extension.

**"Portal ID" is HubSpot vocabulary** — the generic operation here is "find an integration by an entity's external ID." That generic primitive belongs in core; the HubSpot-specific wrapper belongs in the HubSpot api-module's extension.

Concrete proposal:

- **Core** exposes `IntegrationBase.findIntegrationByEntityExternalId(externalId, moduleName?)` and likely a `list*` variant for the ambiguous-but-legitimate case where one externalId maps to multiple integrations (distinct from the throw-on-cross-tenant-routing guard, which would stay).
- **The HubSpot api-module's extension** exposes a thin `findIntegrationByPortalId(portalId)` wrapper that calls the core primitive.

**Why it matters** — a healthy number of api-modules route inbound webhooks by account identifier alone:

- Slack — `team_id`
- Microsoft Teams — `tenant_id`
- Asana — `workspace_id`
- Google Workspace — `workspace_id`
- HubSpot — `portal_id`

Every one of those will want this reverse-lookup. Keeping the generic operation in core (platform-neutral, reusable) and the platform-vocabulary wrapper in each extension (self-documenting for that platform's developers) is the right split. Same rule applies to any helper: if you can name it in a single platform's vocabulary, it belongs in that platform's extension, not in core.

Will pull this rename + relocation into the ADR update on #591 as a worked example of the core / API-module boundary, and add an Open Question about encoding the rule as a contributor guideline.

---
_Generated by [Claude Code](https://claude.ai/code)_

d-klotz:
Pushed 0836cf99 addressing all three threads:

1. **Codex — entity-level ambiguity** ✓
   - Added `findEntities(filter)` to `ModuleRepository` (interface + mongo, postgres, documentdb, legacy)
   - Helper now uses it and throws on entity-level multi-match _before_ the integration check
   - New test `throws on entity-level ambiguity (>1 entity matches externalId)` locks the case

2. **Sean — rename + relocate** ✓
   - `findIntegrationByPortalId` → `findIntegrationByEntityExternalId` (platform-neutral name in core)
   - Added sibling `listIntegrationsByEntityExternalId` for the legitimate one→many fan-out case (returns array, does not throw)
   - `EXTENSIONS.md` now documents the boundary rule with the HubSpot worked example: thin `findIntegrationByPortalId(portalId)` wrapper belongs in `@friggframework/api-module-hubspot/extensions/webhooks`, not in core
   - JSDoc on the core method explicitly calls out where platform-named wrappers belong

3. **Sonar — version disclosure** ✓
   - `app.disable('x-powered-by')` on the test harness in `integration-defined-routers.test.js`

15 unit tests cover the renamed helper + the new list sibling — entity ambiguity, integration ambiguity, orphans, dedup across multi-entity, moduleName forwarding. All green. Existing `integration-defined-routers` suite still passes.

On your other comment re. Queues / Crons / User Actions as extension primitives — agreed those belong on the roadmap, just keeping this PR scoped to webhook/routes/events/workers. Happy to track them as follow-ups against #591 or as a separate issue, whichever you prefer.

seanspeaks:
Task linked: [FRI-498 Fix findIntegrationContextByExternalEntityId to query by entity binding, not user](https://app.clickup.com/t/9013279251/FRI-498)

seanspeaks:
:rocket: PR was released in [`v2.0.0-next.100`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.100) :rocket:

PR #598, opened 2026-06-03
fix(core): discard webhook message when integration deleted mid-flight

d-klotz:
## What

When an integration is **hard-deleted while one of its queue messages is still being processed** (e.g. a user uninstalls mid-flight), the `QueueWorker` now **discards the message instead of retrying it into the DLQ**.

## Why

The worker already hydrates the integration up front and discards if it's gone (`loadIntegrationForWebhook` → null, or the `DISABLED`/`ERROR` status checks). But an integration can be deleted **after** that hydration check, **while the handler is making external API calls** — a time-of-check/time-of-use race.

Once the parent `Integration` row is gone, every subsequent DB write fails. Observed in production: a Pipedrive uninstall during a ~6,050-event webhook burst produced Prisma `P2003` foreign-key violations on `IntegrationMapping_integrationId_fkey` from the mapping upsert. Because a `P2003` carries no `statusCode`, the existing halt-on-4xx logic can't stop it, so the message is retried to `maxReceiveCount` and lands in the DLQ — for what is really an expected lifecycle event. The same race also surfaces as a masked `"Integration not found"` (when the error-recording write itself fails) or a revoked-credential `401`.

A pre-check before the work can't close this race, and translating the specific Prisma error into a typed error is fragile (it can be masked by a downstream error-recording write before any handler sees it).

## How

On the **error path only**, re-check whether the integration still exists; if it's **confirmed gone**, discard the message (no retry) instead of letting it reach the DLQ.

- New `integrationExists(integrationId)` helper — returns `false` only when the lookup confirms the integration is gone (record `null` or a `"not found"` error). A **transient** lookup failure returns `true`, so genuine errors keep their normal retry path.
- Checking **live state** (rather than inspecting the error) is robust to *whatever* error surfaced first — the FK violation, a masked `"not found"`, or a revoked-credential `401` — and covers **every integration** with no per-integration handler code.
- Cost is a single indexed lookup, only on the (rare) error path; the happy path is untouched.

## Tests

`packages/core/handlers/backend-utils.test.js` (new):
- `integrationExists`: found → true, null → false, `"not found"` → false, transient error → true (don't discard on a DB blip).
- `createQueueWorker`: discards (no throw) when the integration is gone after a processing error; re-throws the original error when the integration still exists (so transient failures still retry).

### DIFF
--- a/packages/core/handlers/backend-utils.js
+++ b/packages/core/handlers/backend-utils.js
@@ -110,6 +110,22 @@ const loadIntegrationForWebhook = async (integrationId) => {
     return instance;
 };
 
+// Returns false only when the integration is confirmed gone; a transient
+// lookup failure returns true so genuine errors keep their normal retry path.
+const integrationExists = async (integrationId) => {
+    const integrationRepository = createIntegrationRepository();
+    try {
+        const record =
+            await integrationRepository.findIntegrationById(integrationId);
+        return Boolean(record);
+    } catch (error) {
+        if (error.message?.includes('not found')) {
+            return false;
+        }
+        return true;
+    }
+};
+
 const loadIntegrationForProcess = async (processId, integrationClass) => {
 
     const { processRepository, integrationRepository, moduleRepository } =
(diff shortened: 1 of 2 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.90`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.90) :rocket:

PR #599, opened 2026-06-08
docs(skills): restructure frigg skill into SKILL.md + references

d-klotz:
## What

Restructures `.claude/skills/frigg` from a single ~1500-line `SKILL.md` into a lean overview (~170 lines) plus six domain reference files, applying **progressive disclosure** so Claude only loads the context relevant to the task at hand.

```
.claude/skills/frigg/
├── SKILL.md                         # overview, architecture, integration pattern, CLI, anti-patterns, quick ref
└── references/
    ├── api-modules.md               # module structure, auth requester classes, definitions, JSON Schema forms
    ├── auth-testing.md              # frigg auth: flows, what it tests, saved creds, troubleshooting
    ├── infrastructure.md            # domain builders, scheduler (builder + command API), health, VPC, osls, doctor/repair
    ├── security-encryption.md       # field-level encryption architecture, env config, registry
    ├── management-api.md            # full HTTP endpoint reference for a deployed app
    └── development.md               # fast iteration, TDD, canary, command system, Delegate, debugging
```

## Accuracy fixes (verified against `next`)

- Remove the **nonexistent `frigg search`** command and the **non-functional `frigg init`** scaffold (`frigg-cli/templates/backend` is not shipped, so it crashes with `ENOENT`). Document the working bootstrap path instead (clone an example app, or hand-roll + `frigg install`).
- Correct the **Management UI port** — dev mode serves the Vite frontend on `:5173`, API via `--port` (default `3210`), not `3002`.
- Add the **`doctor`, `repair`, `generate-iam`** CLI commands (were missing).
- Add the **`scheduler` and `health` infrastructure domains** to the builder list.
- Correct the **monorepo layout** — `core`, `database`, `encrypt`, `integrations`, etc. are sibling top-level packages, not nested under `core/`.

## Structure

- Move all "when to use" guidance into the `description` frontmatter (the body only loads *after* the skill triggers, so a body "When to Use" section is dead weight).
- Remove duplicated sections — hexagonal architecture, the golden rules, TDD, and anti-patterns were each stated 2–3 times.

## Preserved

Nothing dropped. The full **Management API endpoint reference** and the **scheduler command API** (`createSchedulerCommands`) are migrated into the new reference files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<details>
  <summary>📦 Published PR as canary version: <code>2.0.0--canary.599.9c0db73.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
[...]

PR #600, opened 2026-06-08
docs(skills): expand frigg Management API reference with auth methods + provisioning workflow

d-klotz:
Follow-up to #599 (now merged). Expands `.claude/skills/frigg/references/management-api.md` with practical auth guidance and an end-to-end provisioning workflow, sourced from an internal `frigg-management-api` guidance skill and generalized to the framework.

## Auth — which method to use

Makes the choice explicit (this is the part that trips people up):

- **Shared secret — `x-frigg` headers** (`x-frigg-api-key`, `x-frigg-appuserid`, `x-frigg-apporgid`): for **backend-to-backend** when there is **no Frigg UI** — server-to-server, scripts, CI/CD, OAuth redirect handlers. The default for backend integrations. Documents `x-frigg-apporgid` (only when `organizationUserRequired: true`) and the `FRIGG_APP_API_KEY` deployment requirement.
- **JWT bearer — user/password**: only when a **Frigg Management UI / end-user accounts** exist (users log in to get a token).

> Rule of thumb: no UI → x-frigg headers; UI with user login → JWT.

## Also added

- Base URL + env setup.
- End-to-end provisioning curl workflow: authorize entity → create integration → trigger `INITIAL_SYNC` → verify, with both header styles shown.
- SKILL.md reference pointer updated to surface the auth distinction.


d-klotz:
Superseded by #601, which splits the `frigg` skill into focused companion skills. The Management API auth + provisioning content from this PR now lives in the new `frigg-management-api` and `frigg-user-actions` skills there.

PR #611, opened 2026-06-29
feat(core): add human-readable labels to integrations report

d-klotz:
## Summary

Adds **human-readable integration labels** to the read-only integrations report (`GET /api/v2/reports/integrations`). The endpoint previously identified each integration only by its `config.type` slug (`"zoho"`, `"attio"`, ...). A downstream consumer (a client-facing "Integration Health" dashboard) needs friendly names ("Zoho CRM", "Pipedrive"). That name already lives on each integration class — this surfaces it.

Two **additive** fields:
- `metrics.byType[].label` — human-readable name per type bucket
- `metrics.typeLabels` — top-level `{ slug → label }` map, so callers can label `integrations[].type` rows without bloating each row

Sourced from each registered integration class's `Definition.display.label`, with a **slug fallback** when no registered class supplies one.

### Note on the field source
The original spec assumed the name lived at `Definition.display.name`. In practice that's a vestigial placeholder (`'Integration Name'`) — real integration classes populate `Definition.display.**label**`, and the framework's own DTO (`integrations/options.js:20`) reads `label`. Reading `display.name` would have made the feature a silent no-op, so this reads `display.label`.

## Design
- **Use case** (`list-integrations-report.js`): accepts an optional injected `typeLabels` map; adds `label: typeLabels[type] || type` to each `byType` bucket and exposes `metrics.typeLabels`. Stays storage-agnostic — all DB adapters (Postgres/Mongo/DocumentDB) benefit with zero adapter changes.
- **Router** (`reporting-router.js`): `buildTypeLabels()` loads registered classes via `loadAppDefinition()` (mirrors `handlers/routers/health.js`), reading `display.label` and filtering the base placeholder. Wrapped in try/catch so reporting never throws if the app definition can't load (→ labels fall back to slugs).

## Compatibility
- `type` unchanged; `label`/`typeLabels` are purely additive.
- `schemaVersion` stays `1` (additive contract).
- No new auth, routes, env vars, or infrastructure. Read-only. No change to `integrations[]` row shape.

## Test plan
- [x] `npx jest reporting` → 6 suites / 35 tests green (use-case label-from-map + slug fallback + `typeLabels` echo; router mocks `loadAppDefinition` and asserts `byType` labels)
- [x] ESLint: 0 errors on touched files
- [x] OpenAPI spec (`docs/frigg-management-api.yml`) updated for `label` + `typeLabels` (YAML validated)
- [x] No consumers of `byType`/`ListIntegrationsReport` outside the reporting module

## Notes
- A reporting `README.md` did not exist on `next`; it's created here.
- Each client Frigg instance must bump `@friggframework/core` + redeploy for labels to appear; the dashboard will render `label || type` separately.

### DIFF
--- a/docs/frigg-management-api.yml
+++ b/docs/frigg-management-api.yml
@@ -906,7 +906,9 @@ components:
                                 Map of `type` slug to human-readable label, for
                                 resolving labels on `integrations[].type` rows.
                                 Contains only types whose registered integration
-                                class supplies a `display.label`.
+                                class supplies a non-default `display.label`
+                                (classes still carrying the IntegrationBase
+                                default `Integration Name` are excluded).
                             additionalProperties:
                                 type: string
                             example:
--- a/docs/frigg-management-api.yml
+++ b/docs/frigg-management-api.yml
@@ -885,12 +885,32 @@ components:
                                     type:
                                         type: string
                                         example: hubspot
+                                    label:
+                                        type: string
+                                        description: >-
+                                            Human-readable name from the
+                                            integration class's
+                                            `Definition.display.label`; falls back
+                                            to the `type` slug when no registered
+                                            class provides one.
+                                        example: HubSpot CRM
                                     total:
                                         type: integer
                                     byStatus:
                                         type: object
                                         additionalProperties:
                                             type: integer
+                        typeLabels:
+                            type: object
+                            description: >-
+                                Map of `type` slug to human-readable label, for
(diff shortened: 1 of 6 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.98`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.98) :rocket:

PR #614, opened 2026-07-03
fix(core): atomic config patches + honest create/delete lifecycle

d-klotz:
## What

Two related fixes to `@friggframework/core`:

**1. Atomic, fail-loud config writes.** `patchIntegrationConfig(id, patch)` — a database-side atomic shallow merge — is added to the repository port and all three adapters (postgres: `config || $1::jsonb` in one `UPDATE`; mongo: `findAndModify` with per-key `$set`; documentdb: per-key `$set` + read-back with the write result checked). `IntegrationBase` gains `patchConfig()` and `updateConfig()`, which throw on failure and keep `this.config` in sync with what was actually persisted. A patch value of `null` is allowed and sets the key to null (the way to clear a field over the HTTP `PATCH` endpoint) — it's a value write, not a key removal; `undefined` is rejected. Removing a key entirely is what the full-replace `updateIntegrationConfig` is for (now also reachable as `this.updateConfig()`).

**2. Two new lifecycle statuses: `IN_CREATION` and `IN_DELETION`.** New integrations are born `IN_CREATION` instead of `ENABLED`. `validateConfig()` is a passive check — it reports whether a required field is missing (recording a warning per missing field) but does not itself change status, mirroring the existing `testAuth`/`reconcileAuthStatus` split. The default `onCreate()` decides between `NEEDS_CONFIG` and `ENABLED` from that result; the default `onUpdate()` merges any submitted config in first, then heals `NEEDS_CONFIG → ENABLED` only (other statuses are left alone — a config check shouldn't silently un-pause `DISABLED` or auto-heal `ERROR`/`IN_DELETION`). `validateConfig()` itself is fixed — it read a legacy array-shaped config-options format that no real integration returns, so it silently did nothing (previously it threw against the real shape, a separate bug). Deletion marks `IN_DELETION` before teardown runs (in the delete use-case, not `onCreate`/`onDelete`, since integrations commonly override the lifecycle hooks and call `super` at the very end — marking it there would be too late). If teardown throws, the integration stays `IN_DELETION` — deletion does not proceed — and an error message is recorded on the record explaining why; the use-case rethrows so the caller sees the failure. The queue worker discards events for `IN_DELETION` integrations, and it's excluded from create-time duplicate matching, so it's inert rather than either a fake-healthy reinstall target or a still-processing "ghost".

## Why

**Config:** `Integration.config` was persisted as a whole-JSON replace, and callers build the next config as `{ ...this.config, ...newKeys }` from an in-memory snapshot. When an install runs two config writers concurrently — for example two webhook-registration legs in the same setup flow — whichever writes second replaces the first writer's keys from a stale snapshot, silently dropping them. This can strand an externally-created resource (e.g. a webhook registered with a third party) whose identifier is now recorded nowhere in config, and can drop a stored signing secret, quietly disabling signature verification. Nothing throws, nothing is logged, and the integration looks fine.

**Lifecycle:** integrations were born `ENABLED`. Any failure between the row insert and the first explicit status write left a permanently `ENABLED` row that never finished setup — indistinguishable from a healthy integration. Symmetrically, a deletion whose teardown fails partway through left no signal that the integration was mid-teardown, so a fresh install for the same user/type/entities could reuse it as if reconnecting to something healthy.

## How

Red-green-refactor throughout. Config: shared patch-validation contract → repository port + in-memory test double → the three adapters → full-replace use case → `IntegrationBase` methods → command facade. Lifecycle: schema defaults + two split Postgres migrations (`ALTER TYPE ... ADD VALUE` can't be used in the same transaction it runs in, so the enum additions and the default-value change are separate migrations) → `onCreate`/`onUpdate`/delete-use-case wiring → worker discard list.

The config merge is **shallow**: only top-level keys are affected, and each key in the patch replaces its existing value wholesale (a nested object is overwritten as a block, not deep-merged).

**This PR went through several rounds of adversarial + human review**, which found and fixed:
- The documentdb adapter discarded the update-command result, so a rejected write (size limit, `writeErrors`) fell through to a read-back that reported success with the patch silently unapplied.
- The "concurrent disjoint patches" regression test passed vacuously against a fully-synchronous test double (no interleaving was possible, so it would have passed against a broken implementation).
- `_findDuplicate` (the create-time dedupe check) treated an `IN_DELETION` row as a live duplicate, handing a reinstall attempt a row nothing ever moves out of `IN_DELETION` — a fake "success" the queue worker then silently discards forever. `IN_DELETION` rows are now excluded from duplicate matching.
- A reused `IN_CREATION` row (a previous create that never finished) was silently returned through the same path as a healthy integration. Not auto-healed — safely retrying setup needs idempotent remote-resource creation first, a separate, larger change — but now logged loudly instead of masquerading as success.
- The create-time **reuse path ignored the caller's config**: reconnecting/retrying create with the same user/type/entities but changed settings kept the old config. `_reuseExisting` now patches the requested config onto the reused row (shallow merge through the repository, so keys added during the original create — webhook ids, secrets — survive).
- **Rejecting `null` in `patchIntegrationConfig`** removed the only way to clear a config field over the HTTP `PATCH` endpoint (the full-replace path isn't exposed over HTTP). `null` is now allowed as a value write (sets the key to null); `undefined` is still rejected.
- Default `onUpdate` called `validateConfig()` but never persisted the submitted config. Fixed to merge it in via `patchConfig` first, then re-validate.
- Both Prisma enum declarations were reordered so `IN_CREATION`/`IN_DELETION` are declared last, matching the physical order Postgres actually produces (`ALTER TYPE ... ADD VALUE` always appends) — avoids a schema/DB drift warning on the next `prisma migrate dev`.

## Verification

All config-path and lifecycle-path suites green (`packages/core/integrations/`, `packages/core/application/commands/`, `packages/core/handlers/`, `packages/core/reporting/`). Remaining failures across the wider suite pre-exist on `next` and are unrelated (encryption/KMS key setup, OAuth2 requester test fixtures, a couple of param-name mismatches) — verified none touch `Integration.status`, config, or the create/update/delete use-cases this PR changes.

## Not addressed here (deliberately out of scope)

- **Retry-safe setup re-run.** A row stuck `IN_CREATION` has no automatic recovery — safely re-firing `ON_CREATE` needs idempotent remote-resource creation (list-before-create) so a retry can't duplicate a webhook the failed attempt already registered. Separate, larger change.
- **Mongo `patchIntegrationConfig` malformed-id handling.** Doesn't guard a malformed ObjectId string before use, so it throws a raw parse error rather than a clean "not found" (postgres and documentdb both guard). Nit — it throws, nothing silent.
- **Nested-value validation.** The shared config-patch validator checks only top-level keys, not nested EJSON-wrapper-shaped values. Not exploitable; on Mongo/DocumentDB such a value is type-coerced rather than stored verbatim, so it's type-lossy only.
- **Rollback ordering.** Once any row is written with `IN_CREATION`/`IN_DELETION`, rolling the app back to a Prisma client that predates this migration will fail to read that row. Noted in the migration files; backfill such rows to a known status before rolling back.


packages/core/integrations/repositories/config-patch-shared.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * A patch is an atomic shallow merge into Integration.config: every key in
+ * the patch is written, every other existing key is left untouched. This
+ * contract is enforced here, before any adapter emits a query, so a bug fix
+ * (or a tighter rule) fixes all three backends in one place.
+ *
+ * Keys may not contain '.' or start with '$' — both are reserved by the
+ * MongoDB/DocumentDB update-command syntax the mongo and documentdb adapters
+ * use to address `config.<key>` paths. Patch values may not be null or
+ * undefined; deleting a key requires a full-replace via updateConfig.
```
d-klotz: [P1] Rejecting `null` here removes the only public way to clear a config field. The management API still only exposes `PATCH /api/integrations/:integrationId`, so clients can no longer delete a key or intentionally set it to `null`; the full-replace `updateConfig` path mentioned here is not available over HTTP.

d-klotz (commented):
Posting the config-clearing comment.

packages/core/integrations/repositories/config-patch-shared.js
d-klotz: [P1] Rejecting `null` here removes the only public way to clear a config field. The management API still only exposes `PATCH /api/integrations/:integrationId`, so clients can no longer delete a key or intentionally set it to `null`; the full-replace `updateConfig` path mentioned here is not available over HTTP.

d-klotz (commented):
Posting the config-clearing comment.

packages/core/integrations/use-cases/create-integration.js
```diff
@@ -114,6 +128,14 @@ class CreateIntegration {
...
 
         const matches = candidates.filter((integration) => {
+            // An IN_DELETION row is either mid-teardown or a zombie left
+            // behind by a failed deleteIntegrationById call — never a live
+            // integration. Treating it as a duplicate would hand a reinstall
+            // attempt a row nothing will ever move out of IN_DELETION,
+            // silently discarded by the queue worker forever. Let a fresh
+            // create proceed instead; the zombie is cleaned up out-of-band,
+            // matching how teardown failures are already handled here.
+            if (integration.status === 'IN_DELETION') return false;
```
d-klotz, line 151: [P1] This duplicate-matching path still ignores the caller's new `config` payload. If a user reconnects or retries create with the same user/type/entity set but changed non-`type` settings, we now reuse the old row unchanged instead of persisting the requested config. Either include the effective config in the dedupe key or patch the reused row before returning it.

d-klotz (commented):
Posting the integration reuse comment.

packages/core/integrations/repositories/config-patch-shared.js
d-klotz: Fixed in 08238ff5. `validateConfigPatch` now allows `null` values — a `null` in the patch sets the key to `null` via the adapters' existing `||` / `$set` writes, so `PATCH /api/integrations/:id` can clear a field again. It's a value-write (the key stays present with a null value, consistent across postgres/mongo/documentdb), not a key removal; removing a key entirely still needs the full-replace path. `undefined` is still rejected — it can't arrive over JSON and signals a programming error.

packages/core/integrations/use-cases/create-integration.js
d-klotz: Fixed in 08238ff5. `_reuseExisting` now patches the caller's `config` onto the reused row before returning it — a shallow merge through the injected repository, so changed settings take effect while keys the original create added (webhook ids, secrets) survive. `type` is re-written to the same value it dedupe-matched on (a no-op).

packages/core/prisma-postgresql/schema.prisma
```diff
@@ -150,7 +150,7 @@ model Integration {
   id     Int               @id @default(autoincrement())
   userId Int?
   user   User?             @relation(fields: [userId], references: [id], onDelete: Cascade)
-  status IntegrationStatus @default(ENABLED)
+  status IntegrationStatus @default(PROCESSING)
```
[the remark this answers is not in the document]
d-klotz: The born status was reworked from `PROCESSING` to `IN_CREATION` after this comment, and the queue-worker guard now discards `IN_DELETION` on both hydration paths. `IN_CREATION` is deliberately *not* discarded: initial-sync `processId` messages are enqueued and must run while the integration is still completing setup — discarding that state would break initial sync. External webhook deliveries can't arrive during `IN_CREATION` anyway, since the vendor-side webhook isn't registered until setup runs. A failed `onCreate` leaves the row `IN_CREATION` (visibly incomplete) and it's excluded from create-time dedupe, so a reinstall creates a fresh row rather than reusing the stuck one.

seanspeaks:
:rocket: PR was released in [`v2.0.0-next.99`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.99) :rocket:

PR #634, opened 2026-08-05
fix(core): record why credentials were rejected before flipping to ERROR

d-klotz:
## Summary

When a module reports `CREDENTIAL_INVALIDATED`, `IntegrationBase.receiveNotification` flips the integration to ERROR via `persistStatus`, which writes **only the status column**. The reason reaches the `console.log` line and nowhere else, so the integration surfaces to end users and support as broken with an empty `errors` array.

This PR records the cause alongside the flip.

## Why

The delegate payload already carries `reason` and `statusCode` — they were logged and discarded. The consequences compound, because integrations in ERROR have their queued webhooks discarded, so the state is both sticky and unexplained.

Measured in one adopter's production database:

- **74 of 333 ERROR integrations (22%) recorded no cause at all** — empty `errors`, `warnings`, `logs` and `info`.
- Accruing **1 to 4 per day**, across five different integration types.
- Every case traced was a genuine expired or revoked authorization, i.e. exactly the condition this delegate exists to report.
- Diagnosing a single one required a CloudWatch Logs Insights query pinned to the second the flip happened.

## What changed

`receiveNotification` now calls `recordCredentialRejection` before `persistStatus('ERROR')`.

It reuses the `'Authentication Error'` title and the wording `testAuth` already writes (`integration-base.js:381`), so the passive auth check and the delegate produce a consistent diagnostic instead of two competing vocabularies for the same condition.

Two deliberate constraints:

**The delegate's `reason` is not persisted.** It is the `FetchError` message, which embeds the serialized request — and `FetchError` blanks that only when `STAGE !== 'dev'`, so outside prod it contains the `Authorization` header. These messages are surfaced to end users, so only the status code crosses over. The full text remains in the log line directly above.

**The write is best-effort and ordered first.** A failed diagnostic write can never prevent the status flip that stops processing on dead credentials; recording before the flip means a failed *status* write still leaves a cause behind.

## Test plan

- [x] 7 new tests in the existing `integration-base-receive-notification.test.js`, written test-first
- [x] Covers: recorded content, ordering guarantee, both failure paths, that a request-echoing `reason` containing an `Authorization` header stays out of the persisted body, and that `CREDENTIAL_VALIDATED` and unhydrated instances record nothing
- [x] `integrations` + `modules` suites: **587 passing**, up from 578 on `next`
- [x] The 11 failures in 3 suites (`integration-mapping-repository-documentdb-encryption`, `find-integration-context-by-external-entity-id`, `oauth-2`) are **pre-existing on `next`** and unrelated — verified by running the same command on a clean checkout
- [x] Prettier and ESLint clean (0 errors)

## Reviewed for correctness and contract

A correctness/contract pass caught a security regression in an earlier revision of this branch and it is fixed in `769f9999`. Recording it here because the reasoning is the interesting part.

The two grant paths originally forwarded their **caught error** to `notify(DLGT_INVALID_AUTH, error)`. But the request bodies there hold a raw `password` (`getTokenFromUsernamePassword`) and a raw `client_secret` (`getTokenFromClientCredentials`), `FetchError` embeds the request body in its `message` whenever `STAGE === 'dev'`, and both `markCredentialsInvalid` and `receiveNotification` log that message. Because these sites previously notified with **no payload**, `markCredentialsInvalid`'s `if (diagnosticInfo)` guard kept the logs empty — so forwarding the error newly sent end-user passwords and client secrets to CloudWatch on any dev-stage deployment. They now forward `{ statusCode }` only, which is all the delegate ever needed, and the tests assert the secret cannot reach the payload rather than merely checking the shape.

Also confirmed by that pass:

- **Ordering is coherent in every interleaving.** Message-written/status-failed leaves ENABLED with an auth error recorded, which is exactly what `testAuth` already does. Record-before-flip is the right order: a crash between the two leaves the diagnosable state rather than the ERROR-with-empty-`errors` state being fixed here.
- **Idempotent under concurrency.** Six firings in eight seconds converge on one message plus status ERROR, because of the clobber behaviour described below. This change does not worsen the pre-existing read-modify-write race.
- **Non-Error payloads degrade safely** through `diagnosticInfo.message ?? diagnosticInfo`.
- **Nothing else persisted can carry a secret** — only the developer-set `moduleName` and a numeric `statusCode`.

## Known adjacent bugs, not addressed here

**`updateIntegrationMessages` replaces rather than appends.** `integration-repository-postgres.js` reads `integration.messages?.[messageType]`, but the `Integration` model has no `messages` column — `errors`/`warnings`/`info`/`logs` are four separate top-level `Json` columns. So the array is always rewritten with a single element. Same in the mongo and documentdb repos.

That is load-bearing here in an unintuitive way: it bounds this write to one entry, which is why a credential storm cannot bloat the row. It also means this entry overwrites any earlier, more specific diagnostic. **Fixing that mismatch should land together with same-title dedupe** — otherwise appends start working and a flip storm grows the array without bound.

## `refreshAuth` now honours its own docstring

`oauth-2.js:293` has always documented "On failure, notifies delegates via `DLGT_INVALID_AUTH`", but the catch only logged and returned `false`. Three tests asserted the documented behaviour and had been red on `next` for as long as they existed. They now pass.

It forwards `{ statusCode }` rather than the caught error, for the same reason as the grant paths above: `refreshAccessToken`'s body carries `client_secret` in a `URLSearchParams` that `FetchError` explicitly stringifies into its message outside prod. A test pins that the secret cannot reach the delegate.

The three stale assertions expected a bare `notify(DLGT_INVALID_AUTH)` with no payload — an expectation that was already wrong for the 401-retry test, which reaches the delegate via `Requester._invalidateAuth` and has been receiving a `FetchError` since that method existed. All three now assert the payload.

The delegate now fires **twice** on the only path that calls `refreshAuth`, because `Requester` also calls `_invalidateAuth` when it returns `false`. The `CREDENTIAL_INVALIDATED` branch is therefore now guarded on the integration already being in ERROR, mirroring the `CREDENTIAL_VALIDATED` branch's existing `if (this.status !== 'ERROR') return`.

That guard is deliberately at the integration level and **not** in `Module.markCredentialsInvalid` on `credential.authIsValid === false`, which is the more obvious place. Two integrations can share one `Credential` row — observed in production, where two Zoho integrations shared credential 12369. A credential-level guard would let the first integration's 401 mark the row invalid and then return early for the second, leaving it ENABLED with a dead credential: exactly the silent-broken state this PR exists to fix. Integration status is per-integration, so the guard belongs there, and a test pins that a second integration sharing the credential still flips.

This removes the sequential duplicate, which is the guaranteed one. It does not deduplicate concurrent invocations — those each hydrate their own instance and read status ENABLED before any write lands — so the observed "six firings in eight seconds" case is damped but not eliminated. Doing that properly needs a conditional update or an advisory lock, which is a bigger change than this PR should carry.

## Test results

Whole `@friggframework/core` package: **1592 passing, up from 1586 on `next`**, and the 12 remaining suite failures are byte-identical to `next` (verified by diffing the failure list against a stashed baseline, not by eyeballing counts). Those are pre-existing encryption/repository suites that need a live database.

### DIFF
--- a/packages/core/integrations/integration-base.js
+++ b/packages/core/integrations/integration-base.js
@@ -398,9 +398,8 @@ class IntegrationBase {
      * @returns {string} A user-facing message.
      */
     _authErrorMessage(moduleName, statusCode) {
-        const entity = moduleName ? `your ${moduleName} Entity` : 'your Entity';
         const status = statusCode ? ` (HTTP ${statusCode})` : '';
-        return `There was an error with ${entity}${status}. Please reconnect/re-authenticate, or reach out to Support for assistance.`;
+        return `There was an error with your ${moduleName} Entity${status}. Please reconnect/re-authenticate, or reach out to Support for assistance.`;
     }
 
     /**
@@ -856,7 +855,7 @@ class IntegrationBase {
         if (!this.id) return;
 
         if (delegateString === 'CREDENTIAL_INVALIDATED') {
-            const moduleName = notifier?.name ?? object?.moduleName;
+            const moduleName = notifier?.name;
             const detail =
                 object?.reason || object?.statusCode
                     ? ` (status ${object?.statusCode ?? '?'}: ${
--- a/packages/core/integrations/integration-base.js
+++ b/packages/core/integrations/integration-base.js
@@ -855,6 +855,8 @@ class IntegrationBase {
         if (!this.id) return;
 
         if (delegateString === 'CREDENTIAL_INVALIDATED') {
+            if (this.status === 'ERROR') return;
+
             const moduleName = notifier?.name;
             const detail =
                 object?.reason || object?.statusCode
--- a/packages/core/integrations/integration-base.js
+++ b/packages/core/integrations/integration-base.js
@@ -861,6 +861,7 @@ class IntegrationBase {
                     this.id
                 } — marking ERROR${detail}`
             );
+            await this.recordCredentialRejection(notifier, object);
             await this.persistStatus('ERROR');
             return;
         }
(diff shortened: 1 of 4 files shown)


seanspeaks:

seanspeaks:
:rocket: PR was released in [`v2.0.0-next.110`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.110) :rocket:

PR #636, opened 2026-08-10
fix(core): single-flight the 401 token refresh in Requester

d-klotz:
## Problem

Sync stages drive records through one api-module instance in **concurrent chunks**. When an access token has lapsed (the overnight case), every request in the chunk 401s at nearly the same moment — and each one then refreshed independently, up to `MAX_AUTH_RETRIES`.

That was actively harmful, not just wasteful. The concurrent refreshes rotate the credential out from under each other — Intuit and others force-expire the previous refresh token — so **the framework manufactures its own `invalid_grant` failures**.

Measured on 5 concurrent 401s against one instance, before this change:

| | Before | After |
|---|---|---|
| `refreshAuth()` calls | **3** | 1 |
| Self-inflicted `invalid_grant` | **2** | 0 |
| Requests rejected (of 5) | **4** | 0 |
| `INVALID_AUTH` notifications | **4** | 0 |

Those 4 notifications flag the credential dead even though a refresh had just succeeded. A single invocation can therefore reproduce the production `invalid_grant` signature with no cross-process race at all.

This was found while investigating a customer incident where an Aspire → QuickBooks integration silently orphaned 95 records over four months.

## Change

**1. Single-flight refresh.** Waiters share the initiator's refresh via `_refreshAuthOnce()`, and only the initiator spends the retry budget.

Two guards come with it:

**Re-entrancy — this one is a deadlock, and it is why the naive fix is not enough.** Subclasses issue their token request through `this._post` (`OAuth2Requester.refreshAccessToken` / `getTokenFromClientCredentials`), which re-enters `_rawRequest`. A 401 from the *token endpoint itself* (`invalid_client` — a revoked or rotated secret) would join the in-flight refresh and await the very promise whose resolution depends on that nested call returning. Circular await: it hangs until the function times out, never flags the credential, and poisons the slot for every later request on the instance. An `AsyncLocalStorage` context marks the refresh flow so a 401 raised inside it is fatal, with no subclass cooperation required.

**Token identity.** A request dispatched before a refresh can have its 401 land *after* that refresh finished, meaning the current token was never tried. `_authGeneration` lets it retry with the new token instead of spending another rotation — and every avoided rotation is one less invalidation of the pair a concurrent holder just minted.

**2. The requester side of ADR-031** (the cross-invocation half). Before a refresh, and again after a definitive rejection, the requester re-reads the credential row through the Module (`DLGT_CREDENTIAL_RELOAD` → `Module.reloadCredential()`) and adopts a newer stored refresh token instead of racing it. The refresh errors are split: only a definitive rejection (400, or 401 for `invalid_client`) can invalidate; a timeout, 429, or 5xx is rethrown as a transport failure and stays retryable. Two counters: `frigg.auth.refresh_race_recovered` and `frigg.auth.refresh_race_lost`.

The decision record for all of the above is **[ADR-042](docs/architecture-decisions/042-in-process-single-flight-token-refresh.md)**, in this PR.

## Review-driven changes

- **The adoption moved into the wrapper.** It was the first line of `OAuth2Requester.refreshAuth()`, so any module that overrides that method erased it — including the incident app's QBO module, plus `airwallex` and `marketo`. It now runs in `_adoptOrRefresh()`, which no module overrides, behind an overridable `_adoptNewerCredential()` hook that the base `Requester` defines as a no-op. `refreshAuth()` keeps a guarded copy for the four modules that call it directly (`frontify`, `netx`, `slack`, `terminus`), so every door stays at one database read.
- **The refresh marker now names its requester.** It carried a bare flag, so a second `Requester` used inside a custom refresh read as in-flow and had its own 401 treated as fatal. The deadlock guard is unchanged: a nested 401 on the same instance still fails fast.
- **`refreshAuth()` no longer reports a failed `client_credentials` token request as success.** `getTokenFromClientCredentials()` resolves to `undefined` on failure, so the refresh counted as successful with no new token and bumped the auth generation for nothing.
- **One generation bump per adoption.** The bump lived in both `_adoptNewerCredential()` and the slot.

## Not in this PR

- The queue-worker fix that stops silently acking work for integrations in `ERROR`. It is independent of the requester, and it is the one default-on behavior change on upgrade — an argued exception to ADR-027 that a maintainer should ratify on its own terms. It ships as its own PR; the implementation is recoverable from this branch's history (`0b28b4ad`).
- ADR-031's option 2, the scheduled proactive refresh script.

## Tests

`requester.concurrent-refresh.test.js`, `oauth-2.credential-reload.test.js`, `module-credential-reload.test.js`. The first four concurrency tests fail on `next` with exactly the signature in the table; the deadlock test hangs before the fix and settles within 1500 ms after.

Also covers the safety properties the retry budget existed to provide, so this cannot silently weaken them: the slot releases for a genuinely later 401, the budget still bounds a pathological upstream that 401s after a "successful" refresh, and all waiters reject when the shared refresh fails. Two tests pin the review findings: a module that overrides `refreshAuth()` still adopts, and a second requester inside the refresh chain is not in the flow.

Touched suites: 95 passed, 2 pre-existing skips.

## Notes for review

- The `.finally()` must stay last in `_refreshAuthOnce()` — the stored promise has to be the one that clears the slot, so it is released before any waiter's continuation resumes.
- Behaviour change worth a second opinion: waiters no longer consume retry budget. Intended, but it makes the budget per-initiator rather than per-request.
- Open, all recorded in ADR-042's Negative section: on a definitive rejection each waiter fires its own `INVALID_AUTH` (N+1 `markCredentialsInvalid` writes); a `refreshAuth()` that does its own HTTP outside `_rawRequest` (e.g. `intuit-oauth`) bypasses the per-attempt timeout, and single-flight concentrates that exposure; the adoption re-read needs `readPreference=primary`, which nothing asserts yet.
- A module that overrides `refreshAuth()` keeps the pre-refresh adoption but forfeits the post-rejection recovery, which still lives in that method's catch.
- Base is merged with `origin/next`.

<details>
  <summary>📦 Published PR as canary version: <code>2.0.0--canary.636.7dc8dca.0</code></summary>
  <br />

  ```bash
  npm install @friggframework/admin-scripts@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/core@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/devtools@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/eslint-config@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/prettier-config@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/schemas@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/serverless-plugin@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/test@2.0.0--canary.636.7dc8dca.0
  npm install @friggframework/ui@2.0.0--canary.636.7dc8dca.0
  # or 
  yarn add @friggframework/admin-scripts@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/core@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/devtools@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/eslint-config@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/prettier-config@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/schemas@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/serverless-plugin@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/test@2.0.0--canary.636.7dc8dca.0
  yarn add @friggframework/ui@2.0.0--canary.636.7dc8dca.0
  ```
</details>

# Version

Published prerelease version: `v2.0.0-next.110`

<details>
  <summary>Changelog</summary>

  #### 🚀 Enhancement
  
  - `@friggframework/core`
    - fix(core): single-flight the 401 token refresh in Requester [#636](https://github.com/friggframework/frigg/pull/636) ([@d-klotz](https://github.com/d-klotz))
  
  #### 🐛 Bug Fix
  
  - `@friggframework/admin-scripts`, `@friggframework/core`, `@friggframework/devtools`, `@friggframework/eslint-config`, `@friggframework/prettier-config`, `@friggframework/schemas`, `@friggframework/serverless-plugin`, `@friggframework/test`, `@friggframework/ui`
    - fix(devtools): opt-in lambda.keepNestedNodeModules ships nested node_modules in Lambda packages [#645](https://github.com/friggframework/frigg/pull/645) ([@d-klotz](https://github.com/d-klotz))
    - fix(core): record why credentials were rejected before flipping to ERROR [#634](https://github.com/friggframework/frigg/pull/634) ([@d-klotz](https://github.com/d-klotz))
  
  #### Authors: 1
  
</details>


seanspeaks (commented):
Read through ADR-042 and the implementation. The ADR is excellent — the before/after measurement table is what makes it reviewable, the scoping against ADR-031 is clean (in-process vs cross-invocation, with the measurement showing why 031's re-read is blind inside one instance), and it's unusually honest about its own open issues. It's also the first ADR written to ADR-014's conventions without being asked, with `README.md` and `adrs.json` both updated. Number 042 is collision-free against everything currently in flight.

Two findings, and they're fixable in the same small change.

---

## 1. The ADR-031 re-read is lost on any `refreshAuth()` override

This is the one I'd want resolved before merge.

The two mechanisms sit at different layers, and only one of them survives subclassing:

| mechanism | lives in | survives an override? |
|---|---|---|
| single-flight slot, `_authGeneration`, refresh-flow marker | `requester.js` — **wraps** the call | ✅ |
| `_adoptNewerCredential()` — 031's pre-refresh DB check | `oauth-2.js` — **first line of** `refreshAuth()` | ❌ |
| `_adoptNewerCredentialWithBackoff()` — 031's post-rejection recovery | `oauth-2.js` — inside `refreshAuth()`'s catch | ❌ |

`requester.js` declares `refreshAuth()` abstract and knows nothing about adoption — no reference to `_adopt*`, `DLGT_CREDENTIAL_RELOAD` or `refresh_token` anywhere in it. So a module that overrides `refreshAuth()` keeps single-flight and loses **both** halves of ADR-031.

What that module then does:

- **Inside one invocation** — still protected. Single-flight is in the wrapper.
- **Across invocations** — fully exposed. It refreshes blind, and on a rotating provider it replays a consumed refresh token and revokes the grant. Precisely the failure ADR-031 exists to prevent.

The reason this isn't hypothetical: **ADR-031 rule 5 names the case.** *"the incident app's QBO override takes no argument and reads `this.refresh_token` directly."* Intuit/QuickBooks is the rotation-exposed provider behind the original incident, and it is documented as overriding the exact method that now carries the check.

### Suggested fix

Hoist the pre-refresh check to where single-flight already lives, as an optional hook:

```js
async _runMarkedRefresh() {
    // ADR-031: never hit the provider if another invocation already won
    if (await this._adoptNewerCredential?.()) {
        this._authGeneration++;
        return true;
    }
    const refreshSucceeded = await refreshContext.run(
        { inRefresh: true, requester: this },
        () => this.refreshAuth()
    );
    if (refreshSucceeded) this._authGeneration++;
    return refreshSucceeded;
}
```

Base `Requester` makes the optional call; only `OAuth2Requester` implements `_adoptNewerCredential`. Non-OAuth requesters have no refresh token and simply don't define it, so it no-ops.

Beyond override-immunity, it's the more honest layering: *"re-read before refreshing"* is a property of **being about to refresh**, not of one implementation of refreshing.

The post-rejection backoff is harder to hoist — it's entangled with `_isDefinitiveAuthRejection` inside the catch. Pragmatic split: hoist the pre-refresh check now, and document that overriding `refreshAuth()` forfeits the recovery half.

**Sizing caveat:** I can't see the api-module-library from here, so I don't know how many of the ~46 modules actually override `refreshAuth()`. ADR-031 says 35 are rotation-exposed and cites QBO as an override. That count is what decides whether this is a blocker or a fast follow — worth checking before merge.

---

## 2. The refresh-flow marker has no requester identity

You already flagged this in the ADR's Negative consequences as "an open review finding. The change is small." Confirming it against the shipped code, and noting it lands in the same function as fix 1:

```js
_runMarkedRefresh() {
    await refreshContext.run({ inRefresh: true },            // no identity
                             () => this.refreshAuth());
}
_isInsideRefreshFlow() {
    return Boolean(refreshContext.getStore()?.inRefresh);    // any requester matches
}
```

`refreshContext` is module-level, so it's shared by every `Requester` in the process. Any requester whose call happens inside that async chain reads `inRefresh: true` and treats its own 401 as fatal. A custom `refreshAuth()` that reaches through a *different* `Requester` — and the ADR notes `intuit-oauth` does its own HTTP — gets a healthy credential invalidated instead of refreshed.

The `{ inRefresh: true, requester: this }` in the snippet above covers it, with:

```js
_isInsideRefreshFlow() {
    return refreshContext.getStore()?.requester === this;
}
```

Worth a test that asserts a second requester inside the refresh chain does *not* read as in-flow — it's the kind of thing that silently regresses.

---

## Not blocking

The other two open issues in the ADR both read as fair follow-ups rather than merge blockers:

- **N+1 `markCredentialsInvalid` writes** on a definitive rejection. Not a regression, and the safe direction.
- **`refreshAuth()` doing HTTP outside `_rawRequest` bypasses the per-attempt timeout**, which single-flight concentrates — a hung token endpoint now pins the initiator *and* every waiter until the Lambda times out. Worth its own issue; wrapping `refreshAuth()` in a timeout is the obvious follow-up.

And one you assign to ADR-031 that I'd keep visible: the adoption check reads the credential row, and **a replica-set secondary read can miss the winner's write**. Nothing here asserts `readPreference=primary`, so the re-read can return a false "nothing newer" exactly when it matters most.

d-klotz:
Both fixed in 4df717a9.

**1. Adoption on overrides.** Moved out of `refreshAuth()` into `_adoptOrRefresh()` in the wrapper, behind an `_adoptNewerCredential()` hook that base `Requester` declares as a no-op. Guarded copy stays in `refreshAuth()` for the direct callers, so every path still does exactly one DB read.

Sizing you asked for, counted against the library (55 modules): 3 replace `refreshAuth()` — airwallex, marketo (both needs-updating), plus the incident QBO module. 9 override `refreshAccessToken()`, unaffected. 4 call the stock method directly — frontify, netx, slack, terminus. So: fast follow for the library, real fix for the incident app.

**2. Marker identity.** `{ requester: this }` in the store, identity compare in `_isInsideRefreshFlow()`. Both new tests go red on the previous head.

Post-rejection backoff stays in the catch, as you suggested — overriding `refreshAuth()` forfeits that half. Documented in ADR-042.

One worth flagging: adoption writes only the Frigg fields, so an intuit-oauth/jsforce client stays stale after it. The hook is overridable for exactly that — call super, copy into the client.

Your three non-blockers are unchanged and listed in ADR-042's Negative section.

seanspeaks:
:rocket: PR was released in [`v2.0.0-next.110`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.110) :rocket:

PR #637, opened 2026-08-11
ADR-031: Concurrent OAuth Credential Refresh Across Lambda Invocations

d-klotz:
## What this is

A Proposed ADR. It decides how Frigg handles concurrent OAuth credential refresh across separate Lambda invocations. It needs maintainer ratification, not merge-on-green.

## Why now

A production Frigg app (Aspire → QBO) silently orphaned 95 records over 4 months. Verified root cause, with log and code evidence in the ADR's Context section:

1. Two Lambda invocations refreshed the same QBO credential 368ms apart. The loser presented an already-rotated refresh token and got `invalid_grant`.
2. The loser then flagged the shared credential invalid → integration `ERROR` → the queue worker silently acked every later message. No DLQ entry. That amplifier, not the race itself, is what turned 3 failed invoices into 95 lost records.

PR #636 (per-instance single-flight refresh) ships the in-process half. It cannot reach a race between invocations — this ADR is the cross-invocation half.

## Prior art check

No existing ADR covers refresh concurrency. Nearest: ADR-005 (token refresh as an admin utility example), ADR-006 (credential endpoints), ADR-009 (flags refresh-on-401 as a low-coverage test gap). ADR-003 does not constrain this — it is scoped to the local-dev GUI.

## The decision, in one paragraph

A layered stack, no new AWS infrastructure: (1) failure containment — a lost race is retryable, only genuinely dead credentials invalidate, and the worker's `ERROR` path stops silently acking (with a kill switch); (2) loser recovery — re-read the credential on 401/`invalid_grant` and adopt the winner's token, with an escalation rule when adoption fails; (3) compare-and-swap on the credential write via a plaintext `tokenVersion`; (4) core-owned token expiry plus a narrow pre-flight gate; (5) declared per-module rotation semantics; (6) optional jitter. Provable serialization (SQS FIFO per credential) is deferred behind a named metric (`frigg.auth.refresh_race_lost`).

## The three designs you asked about

All three are evaluated in Alternatives Considered, on mechanism:

- **Central token service / proxy** — rejected as the correctness mechanism (traced: at concurrency > 1 it closes 0% of the observed window; its own outage reproduces the silent-loss chain fleet-wide). Kept alive as a possible later observability topology.
- **DB-event / SNS subscription** — rejected. No SNS protocol can deliver into a running invocation; the implementable form degenerates to a re-read with extra infrastructure and a connection-leak trap (Frigg has no teardown path).
- **Per-API flag + re-read on 401 + dual-write cache** — the re-read is adopted as layer 2 and made unconditional (the flag would leave 35 rotation-exposed modules broken until 35 authors opt in). The dual-write cache is rejected: no write ordering is both useful and safe, and the datastore already is the shared cache.

## What ratification requires

Two named calls, in the open-question section:

1. Is "no failure is silent, losers always recover" a sufficient contract, or does Frigg need provable serialization (layer 7)?
2. Layer 1 is default-on, which is an argued exception to ADR-027's no-behavior-change-on-upgrade bar.

## Review notes

- Numbered **031**: 028–030 are claimed by ADR drafts on unmerged branches (`claude/frigg-adr-ssm-env-management`, `claude/frigg-adrs-app-init-skills-pipeline`).
- The document went through two adversarial review passes; every file:line citation was verified against source, and the register files (README table, `website/roadmap/data/adrs.json`) are updated.
- Unverified items are labelled in-text (Intuit's exact rotation interval/grace window, per-provider grant-revocation behavior, AWS latency estimates).

### DIFF
--- a/docs/architecture-decisions/031-concurrent-oauth-credential-refresh.md
+++ b/docs/architecture-decisions/031-concurrent-oauth-credential-refresh.md
@@ -6,18 +6,46 @@
 
 ## Context
 
-A production Frigg app (Aspire → QuickBooks Online, AWS Lambda, MongoDB)
-silently orphaned **95 records over 4 months**. The trigger is a
-cross-invocation credential rotation race; the reason it stayed silent for
-four months is a separate amplifier in the invalidation and queue-worker
-paths. Both are framework-level, both are reachable by every one of the ~46
-api-modules in the api-module-library, and neither is addressed by the
-per-instance single-flight refresh that already shipped in PR #636.
+This document is written in ASD-STE100 Simplified Technical English.
+
+### Terms used in this document
+
+- **Invocation** means one run of an AWS Lambda function.
+- **Credential record** means the database row that holds the tokens for one
+  connection.
+- **Access token** means a short-life token. The app sends it with each API
+  request.
+- **Refresh token** means a long-life token. The app uses it to get a new
+  access token.
+- **Rotation** means the provider issues a new refresh token and kills the
+  old one.
+- **Race** means two invocations do the same operation at almost the same
+  time. One wins. One loses.
+- **401** means the HTTP status code for "not authorized".
+- **`invalid_grant`** means the OAuth error for a rejected refresh token.
+- **Ack** means the worker tells SQS that a message is complete. SQS then
+  deletes the message.
+- **DLQ** means dead-letter queue. SQS moves a message there after too many
+  failed tries.
+- **Hydrate** means load data from the database into memory.
+- **CAS** means compare-and-swap: a write that succeeds only when a stored
+  value matches an expected value.
+- **Pre-flight check** means a check that runs before a request.
+
+### The problem
+
+A production Frigg app (Aspire to QuickBooks Online, AWS Lambda, MongoDB)
+silently orphaned **95 records over 4 months**. A cross-invocation credential
+rotation race caused the failures. A separate amplifier in the invalidation
+and queue-worker paths kept the failures silent. Both faults are in the
+framework. Both faults can reach each of the ~46 api-modules in the
+api-module-library. PR #636 (the per-instance single-flight refresh) does not
+fix either fault.
 
 ### The race, as observed
 
-Two *different* Lambda requestIds, 368ms apart, both refreshing the same QBO
-credential on 2026-08-09:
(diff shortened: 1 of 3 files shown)


seanspeaks (changes requested):
ADR is too verbose by 60%+ (rough estimate)

The problem can be stated pretty succinctly... Frigg's API modules and requester approach currently does not have any way to deconflict when parallel processes run into auth refresh errors for any OAuth or session based auth. That leads to multiple parallel processes refreshing credentials in-memory and then whichever was the "last write" wins in the DB.

In most cases, this is a non-issue when the API provider allows for either multiple live access tokens, or multi use for a given refresh token.

In some cases, it's an acute issue. Either generating an access token invalidates all prior ones; or refresh tokens are single use.

^^ we can get into more details by pointing to the parts of the code where this happens or doesn't happen, ultimately in the Requester class and the api class.

From this we have some options.

1. We can have each invocation/process/thread "listen" for changes to the specific database table/models upon connection. Any changes will then propagate to the in-process instance of the api class, swapping out or referencing the newly refreshed credential details
2. We can proactively refresh using the new admin scripts with optional core scripts that can run on a cron
3. We can pre-flight freshly fetch a credential, either naively (every time), or in a specific context (i.e. when it's known this is a parallel request, a flag gets added/passed into the api class instance)
4. We can reactively check against the DB first before we attempt an auth refresh or retry (on 401). 

Option 1 feels like it adds too much overhead and may be brittle. Alternatively, it's a good pattern broadly to consider and maybe use in Frigg with distributed workloads.

Option 2 seems like we should build it anyway and have it be automatically installed or offered to be installed via CLI whenever someone installs an API module that has expiritng OAuth tokens with refresh. It won't solve everything but it likely would make the problem more rare for a low cost.

Option 3 feels like we'd incur too much overhead for each request, essentially flooding requests to the DB.

Option 4 feels like the right answer, with the caveat noted below.

1, 3 and 4 also would need to consider delays to the DB read/writes... i.e. maybe it takes an extra 2-5 seconds for a db write to percolate? Depending on how heavily used the DB is?


Verdict/recommendation is Option 2 and 4

From there we can dig into the architecture and considerations.

seanspeaks:

seanspeaks:

seanspeaks:

seanspeaks:

PR #638, opened 2026-08-12
fix(core): write derived process metrics atomically to stop lost counter updates

d-klotz:
## Problem

`UpdateProcessMetrics` writes twice per call. Phase 1 is atomic (`applyProcessUpdate`, single `jsonb_set` UPDATE). Phase 2 then persists derived fields by writing the **whole** `context`/`results` objects back through the legacy `update()` — a blind full-column overwrite. Under concurrent callers, a stale phase-2 write replays its phase-1 snapshot over the row and permanently discards increments other callers landed in between.

This wedged a production sync (HubSpot↔Clockwork, 88 concurrent batch handlers): `processedRecords` ended exactly one 25-record chunk short of `totalFetched`, the completion gate `processed >= totalFetched` never fired, the queue drained, and the process sat in `PROCESSING_BATCHES` forever — with zero errors anywhere (SQS sent == deleted, DLQ empty, Lambda errors 0).

Reproduced against real Postgres: an 88-way concurrent burst lost **1025 of 2200** increments with the current code, **0** with this fix.

## Fix

Phase 2 now writes only the three derived paths it owns (`results.aggregateData.duration`, `results.aggregateData.recordsPerSecond`, `context.estimatedCompletion`) through the same atomic `applyProcessUpdate({ set })` primitive. Derived values stay best-effort under concurrency — same semantics as before — they just can no longer destroy counters. Non-fatal error handling, return value, and websocket broadcast are unchanged.

All three repository backends already support the `set` op: postgres (`wrapSet` + `ensureParents`), mongo and documentdb (`$set` dot paths).

## Tests

- New regression suite: deterministic stale-writer interleave (fails on old code with exactly one chunk's delta lost, the production signature), legacy `update()` never called, derived fields persisted, `estimatedCompletion` omitted when `totalRecords` unknown.
- 4 existing tests retargeted from the old contract (full-blob `update()`) to the scoped set payload; same values asserted.
- `update-process-metrics.test.js`: 26/26. `integrations/use-cases`: 117 passed / 0 failed. Full core suite: identical failure set before/after the change (63 pre-existing, encryption/JWT related, verified by stash + re-run).

## Caveats

- Validated end-to-end on postgres (real DB, deterministic interleave + 88-way burst, exact counters). Mongo/documentdb validated by code inspection only.
- Downstream consumers that pinned behavior on `update()` receiving full blobs would be affected — none found in this repo.

# Version

Published prerelease version: `v2.0.0-next.107`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  - `@friggframework/core`
    - fix(core): write derived process metrics atomically to stop lost counter updates [#638](https://github.com/friggframework/frigg/pull/638) ([@d-klotz](https://github.com/d-klotz))
  - `@friggframework/admin-scripts`, `@friggframework/core`, `@friggframework/devtools`, `@friggframework/eslint-config`, `@friggframework/prettier-config`, `@friggframework/schemas`, `@friggframework/serverless-plugin`, `@friggframework/test`, `@friggframework/ui`
    - fix(core): record why credentials were rejected before flipping to ERROR [#634](https://github.com/friggframework/frigg/pull/634) ([@d-klotz](https://github.com/d-klotz))
  
  #### Authors: 1
  
</details>

### DIFF
--- a/packages/core/integrations/use-cases/update-process-metrics.js
+++ b/packages/core/integrations/use-cases/update-process-metrics.js
@@ -16,11 +16,13 @@
  *
  *   2. Derived-fields phase — duration, recordsPerSecond,
  *      estimatedCompletion. Computed from the post-atomic snapshot and
- *      written via the legacy (non-atomic) `update()` method.
- *      Intentionally best-effort: under concurrent writers they reflect
- *      "whichever handler wrote last" — the same semantics they had
- *      before and all they've ever guaranteed. Preserved for backward
- *      compatibility with consumers (UI, WebSocket listeners).
+ *      written through the same atomic primitive, scoped to only those
+ *      three paths. The values themselves stay best-effort (under
+ *      concurrent writers they reflect "whichever handler wrote last"),
+ *      but the write can no longer clobber another caller's counters.
+ *      Writing the full context/results blob here — as the legacy
+ *      `update()` path did — replayed phase 1's snapshot over the row and
+ *      permanently discarded increments landed in between.
  *
  * Optionally broadcasts progress via WebSocket service if provided.
  *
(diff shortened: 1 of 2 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.107`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.107) :rocket:

PR #642, opened 2026-08-21
fix(deps): clear all production dependency vulnerabilities

d-klotz:
## Summary

`npm audit` on `next` reported 68 advisories (3 critical / 34 high / 27 moderate / 4 low). Splitting dev from production showed the published surface accounted for only 10 of them, and every one had a patch available **inside its existing major version** — so no breaking changes were needed to clear it.

| | before | after |
|---|---|---|
| Production deps (`npm audit --omit=dev`) | 10 (9 high, 1 low) | **0** |
| Full tree | 68 (3 crit, 34 high, 27 mod, 4 low) | 59 (2 crit, 27 high, 27 mod, 3 low) |

## Root cause

Several entries in the root `overrides` block had gone **stale**. Each pin was correct when it was added, but later advisories widened the vulnerable range past it, and the pins were never revisited.

The sharpest case: `velocityjs` was pinned at exactly `2.1.6` while the critical RCE advisory ([GHSA-7gfh-x38p-prh3](https://github.com/advisories/GHSA-7gfh-x38p-prh3)) covers `<=2.1.6`. The override was holding the tree *on* a vulnerable version rather than off it.

### Stale pins bumped

| package | was | now | advisory range |
|---|---|---|---|
| `fast-uri` | 3.1.2 | 3.1.5 | `<=3.1.4` |
| `ip-address` | 10.2.0 | 10.5.0 | `<=10.3.0` |
| `velocityjs` | 2.1.6 | 2.1.7 | `<=2.1.6` (critical RCE) |
| `js-yaml` (scoped) | 4.2.0 | 4.3.1 | `4.0.0 - 4.3.0` |

### New pins

| package | pinned to | advisory range |
|---|---|---|
| `nanoid` | 3.3.18 | `<=3.3.17` |
| `postcss` | 8.5.26 | `<=8.5.22` |
| `body-parser` | 1.20.6 | `<1.20.6` |
| `js-yaml@3` | 3.15.1 | `<=3.15.0` |
| `js-yaml@4` | 4.3.1 | `4.0.0 - 4.3.0` |
| `brace-expansion@1` | 1.1.18 | `<=1.1.17` |
| `brace-expansion@2` | 2.1.4 | `2.0.0 - 2.1.3` |
| `undici@6` | 6.28.0 | `<=6.27.0` |
| `undici@7` | 7.29.0 | `7.0.0 - 7.28.0` |

## Notes on approach

**`js-yaml` and `brace-expansion` are pinned per major line, not globally.** Both have multiple majors live in the tree. A blanket pin would have forced js-yaml 3.x consumers onto v4 (which dropped `safeLoad`) and *downgraded* brace-expansion 5.x consumers to 2.x. The `name@major` override key form keeps each line on its own patched release.

**`ajv`, `socks` and `@istanbuljs/load-nyc-config`** were flagged only for depending on vulnerable `fast-uri` / `ip-address` / `js-yaml`. They clear without a direct change.

**`website/`** has its own lockfile outside the workspaces, so the root audit never covered it. Checked separately — already audits clean, no change needed.

## Test plan

- [x] `npm ci` exits 0 from the regenerated lockfile (2425 packages) — confirms `package.json` and `package-lock.json` are in sync
- [x] All on-disk resolved versions match intent (`fast-uri` 3.1.5, `nanoid` 3.3.18, `postcss` 8.5.26, `body-parser` 1.20.6, `js-yaml` 3.15.1 + 4.3.1, `brace-expansion` 1.1.18 + 2.1.4, `ip-address` 10.5.0, `undici` 6.28.0 + 7.29.0)
- [x] Lockfile diff reviewed: 33 changed entries, all upward, zero downgrades
- [x] `npm audit --omit=dev` → `found 0 vulnerabilities` against the real installed tree
- [x] `admin-scripts` 278/278, `schemas` 98/98, `serverless-plugin` 43/43 passing
- [x] `devtools` results byte-identical to a baseline run on unmodified `next` (32 failed suites / 138 failed tests both before and after) — those failures are pre-existing and unrelated to dependencies

## Out of scope

The remaining 59 advisories are all in build/release tooling and sit behind five major upgrades, each of which is breaking and needs its own testing:

| upgrade | advisories cleared |
|---|---|
| `lerna` 8.1.9 → 10.0.1 | 14 |
| `serverless-offline` → 14.8.0 | 6 (incl. both remaining criticals) |
| `prisma` → 6.12.0 | 3 |
| `nx` 20.3.2 → 23.1.1 | 2 |
| `vite` → 8.2.2 | 2 |

One of these is **not** a version bump: `decompress` has no fixed version at any release, so it needs replacing or dropping rather than upgrading. It reaches the tree through the serverless deploy tooling.

## Suggested follow-up

Production-only audit is now at zero, which makes it a cheap and achievable CI gate. Failing a build on `npm audit --omit=dev` would catch override drift like this automatically, instead of it surfacing months later.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SdchVmNyYEzYD9rBfRugko)_

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2145,9 +2145,9 @@
       }
     },
     "node_modules/@effect/platform-node/node_modules/undici": {
-      "version": "6.27.0",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-6.27.0.tgz",
-      "integrity": "sha512-YmfV3YnEDzXRC5lZ2jWtWWHKGUm1zIt8AhesR1tens+HTNv+YZlN/dp6G727LOvMJ8xjP9Be7Y2Sdr96LDm+pg==",
+      "version": "6.28.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-6.28.0.tgz",
+      "integrity": "sha512-LIY910g9TI13YS95lrMFrs8Rm/u/irgHeTWoKCoteeJ04CUJ92eEfj0rVn+7VKMPBpUPiUoBKfhNyLI23EE/KA==",
       "dev": true,
       "license": "MIT",
       "engines": {
@@ -6646,9 +6646,9 @@
       }
     },
     "node_modules/@npmcli/arborist/node_modules/brace-expansion": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.1.tgz",
-      "integrity": "sha512-WR1cURNjuvBLMZBMbqM0UoE+WAfdUcEV1ccD8PVBVOI+Z3ND4+SZbN8RsfT2bMuG1qwz5RFvPukSZm5fF2D5eA==",
+      "version": "2.1.4",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.4.tgz",
+      "integrity": "sha512-hGfVzPxthbf3+2yjg/RBs60cB0FhqBS/zvdV/4wn4/BmN0bNMMHPc4V/BbFieqf1TKAGGAHnY4eSjajCl0f2Xg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


seanspeaks:

PR #645, opened 2026-09-03
fix(devtools): opt-in lambda.keepNestedNodeModules ships nested node_modules in Lambda packages

d-klotz:
## Summary

Every Lambda package excludes `node_modules/**/node_modules/**`. npm nests a package only when the root copy cannot satisfy the dependant, so the exclusion makes packages resolve whatever is hoisted, and it drops a package entirely when the root copy is a dev dependency. The first `require` then fails at Lambda init with `Runtime.ImportModuleError`.

Seen today on the clockwork-recruiting-integrations dev stack: jsforce 3.10.24 needs undici 8, npm nested it under jsforce because serverless-esbuild and osls pin undici 6 and 7 at the root, the package dropped it, and the `auth`, `hubspot` and `salesforce` functions all died at init for 33 minutes. `readable-stream` under `multistream` fails the same way. Earlier occurrences in the same app were worked around by declaring `js-yaml`, `whatwg-url` and `uuid` as root dependencies.

This PR adds `lambda.keepNestedNodeModules: true`. Default `false`, no behavior change for existing apps (ADR-027's no-behavior-change-on-upgrade bar). When set, function packages ship nested node_modules as npm resolved them, except nested copies of `@friggframework/*`, `aws-sdk`, `@aws-sdk/*` and, with the Prisma layer, `@prisma/*` and `.prisma`: the app's single core, the Lambda runtime and the layer provide those.

### Changes

- `infrastructure/domains/shared/utilities/nested-node-modules.js`: the option read and the pattern list.
- Applied at the five sites that emitted the blanket pattern: base-definition-factory (auth, user, health), integration-builder (integration functions, queue workers, DLQ processor), migration-builder (worker and router), admin-script-builder (executor and router). The app definition is threaded into `createFunctionPackageConfig` and `createFunctionDefinitions`.
- `infrastructure/README.md`: option documented next to `usePrismaLambdaLayer`.

## Test plan

- [x] Unit tests for the helper and for each builder: default keeps the blanket exclusion, opt-in removes it and keeps the scoped ones.
- [x] Full devtools suite.
- [x] Canary `2.0.0--canary.645.2234921.0` deployed to the clockwork dev stack with the flag set and the undici and readable-stream root declarations removed (lockfile restored to the nested placements): every Lambda boots, see below.

### Canary verification (clockwork-recruiting-integrations, dev, 2026-09-03 23:33Z)

Before the fix the same lockfile killed `auth`, `hubspot` and `salesforce__clockwork` at init with `Runtime.ImportModuleError: Cannot find module 'undici'`. With the canary devtools and `lambda.keepNestedNodeModules: true`:

| Function | Probe | Init |
|---|---|---|
| auth | 401 on `/api/integrations` | 1.38 s, no errors |
| hubspot | 404 on an unknown route | 1.42 s, no errors |
| salesforce__clockwork | 200 on the webhook POST | 1.28 s, no errors |

Deployed `salesforce__clockwork` zip: 26 nested `node_modules` directories shipped, including `jsforce/node_modules/undici` (214 files) and `multistream/node_modules/readable-stream` (25 files); 0 nested `@friggframework` files; 0 `@aws-sdk` files anywhere; only the tracked `.env.default` and `.env.example`. Size 37 → 42 MB compressed, 138 → 145 MB unzipped, about 183 MB with the Prisma layer against the 250 MB limit.

Local reproduction of the packaging rule on a production-only install of that lockfile: stripping every nested `node_modules` fails to load the app definition (`Cannot find module 'undici'`); stripping only nested `@friggframework`, `aws-sdk`, `@aws-sdk`, `@prisma` and `.prisma` loads it, with jsforce resolving `node_modules/jsforce/node_modules/undici`.

Frigg CI: the `build (22.x)` failures are in `cloudformation-discovery.test.js` and `vpc-builder.test.js` plus the 10-minute step timeout. Neither suite imports a file changed here, both fail identically on a local checkout of this branch, and the last four CI runs on `next` itself are red. The five suites covering this change pass (169 tests).

### DIFF
--- a/packages/devtools/infrastructure/README.md
+++ b/packages/devtools/infrastructure/README.md
@@ -285,6 +285,24 @@ const appDefinition = {
 
 > ℹ️ When `usePrismaLambdaLayer: false`, Prisma stays inside each bundle and the runtime automatically loads the correct binary. No extra configuration is required.
 
+**Keeping nested node_modules in Lambda packages:**
+
+By default every function package excludes `node_modules/**/node_modules/**`. npm only nests a package when the root copy cannot satisfy a dependant, so the exclusion makes a package resolve whatever version is hoisted, and it removes the package entirely when the root copy is a dev dependency. The first `require` of such a package then fails at Lambda init with `Runtime.ImportModuleError`.
+
+To ship the dependency tree the way npm resolved it, opt in per app:
+
+```javascript
+const appDefinition = {
+    name: 'my-app',
+    lambda: {
+        keepNestedNodeModules: true,
+    },
+    integrations: [{ Definition: { name: 'salesforce' } }],
+};
+```
+
+Nested copies of `@friggframework/*`, `aws-sdk`, `@aws-sdk/*` and, when the Prisma layer is enabled, `@prisma/*` and `.prisma` stay excluded: the app's single core, the Lambda runtime and the layer provide them. Expect each function package to grow by the size of the nested directories that remain (a few MB compressed in a typical app). The default stays unchanged for apps that do not set the flag.
+
 ## Usage Examples
 
 ### Basic Deployment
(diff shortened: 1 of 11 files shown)


seanspeaks:
:rocket: PR was released in [`v2.0.0-next.109`](https://github.com/friggframework/frigg/releases/tag/v2.0.0-next.109) :rocket:

Issue #647, opened 2026-09-17
Define the integration versioning contract (ADR-013 dependency, never authored)


seanspeaks:
## Summary

There is no defined contract for how **integration versions** are declared, compared, or gated. [ADR-013: Integration Version Migrations](https://github.com/friggframework/frigg/blob/next/docs/architecture-decisions/013-integration-version-migrations.md) established the *migration* concept and its execution home, but explicitly deferred the versioning model itself:

> This ADR establishes the migration **concept and execution home**. It deliberately does **not** define how integration versions themselves are declared, compared, gated, or made backward/forward-compatible (`supportedVersions` semantics, semver policy, when a migration is *required* vs *optional*, compatibility windows). That **integration versioning contract** is a separate ADR, to be authored independently.

That separate ADR has never been written. ADR-013 records the consequence in its own Negative section: *"Depends on the forthcoming versioning ADR for its version contract; partial until that lands."*

Today `supportedVersions` is ad hoc, and `packages/devtools/migrations` validates against it with no shared semantics behind the comparison.

**This issue is to track the problem, not to schedule the work.** Filing it so the dependency stops being invisible.

## Why it is getting more expensive to defer

Four separate design efforts now key off a contract that does not exist:

| Depends on it for | What breaks without it |
|---|---|
| **ADR-013** (Proposed) | A migrator is keyed by `(integrationType, fromVersion, toVersion)` with no defined meaning for those versions or for when a migration is required |
| **Sync module revitalization** (in draft) | Canonical sync objects become shared vocabulary across integrations; a breaking change to a shared object ripples into every consumer with no compatibility rule |
| **Canonical models** (in draft) | Forking is the supported path by design, so every upstream schema change is potentially breaking for an unknown number of adopter forks |
| **Orchestrator/worker pattern** (in draft) | State machine definitions change while instances are *in flight*. Instances must be locked to the version they started under, with explicit migrations for state values and context shape — versioning a **running** thing, not a stored record |

That last one is a genuinely different problem from the other three and may not fit the same contract.

## The thorny questions

1. **What is versioned?** The integration type's definition (config shape, mappings, behaviour)? The API module it composes? Both, independently? A module bump that changes nothing for the integration should not force a migration.
2. **How is a version declared and compared?** Semver, and if so what counts as breaking — config shape, mapping shape, or observable behaviour? What are `supportedVersions` semantics beyond "is in this list"?
[...]

PR #620, opened 2026-07-04
docs(adr): draft ADR-030 Integration Versioning (exploratory)

seanspeaks:
## Summary

Partner ADR to [ADR-013 Integration Version Migrations](./docs/architecture-decisions/013-integration-version-migrations.md) (merged in [#617](https://github.com/friggframework/frigg/pull/617)), which explicitly defers the version contract to a separate ADR. This is that ADR.

**Renumbered from ADR-027 to ADR-030** after `027-ssm-parameter-offload-and-env-scoping.md` merged to `next`. This matches the assignment in [#644](https://github.com/friggframework/frigg/pull/644). Conflicts with `next` are resolved on this branch: the SSM row stays at 027, Integration Versioning lands as 030, and ADR-031 / ADR-042 are preserved.

**Deliberately exploratory rather than deciding.** Per Sean's direction, global-routing behavior in serverless environments needs more research before any path is chosen. This ADR lays out the design space, four explored paths with tradeoffs, serverless-specific concerns, and a 12-item edge-case catalog. Path selection is deferred to a follow-up ADR (or a revision of this one) once the open research questions resolve.

## Structure

- **Current state** (grep-verified in `next`): `Definition.version` is a label. Nothing in `packages/core` gates behavior on it. `supportedVersions` is declared but unread.
- **Four dimensions**: declaration, routing, storage, compatibility, and why they couple.
- **Four explored paths**:
  - **Path A**: bump-in-place with migration on read
  - **Path B**: side-by-side classes with per-version routing
  - **Path C**: version selector on `appDefinition`
  - **Path D**: version as a routing dimension outside the class
- **Serverless-specific concerns**: Lambda module-scope cache, edge vs origin routing, signed webhook URL survival, in-flight SQS jobs, EventBridge Scheduler drift, rollback semantics, encryption schema, deploy-time coordination with ADR-012 / ADR-013.
- **12-item edge-case catalog**: mixed-version fleets, third-party-forced bumps, adopter calls to old URLs, credential/mapping/config shape changes, beta versions, multi-tenant pinning, cross-integration references, and more.
- **8 open questions + 4 research follow-ups** Sean called out (global routing, signed URLs, Lambda aliases + weighted traffic, industry examples).

## Placement

`docs/architecture-decisions/030-integration-versioning.md` in the consolidated numbered register (per ADR-014), directly cross-referencing ADR-013.

## Test plan

[...]

PR #648, opened 2026-09-17
docs(adr): 044-047 — sync engine revitalization, capability traits, canonical models, orchestrator/worker

seanspeaks:
Four connected ADRs. They can be read and reviewed independently, but 045 and 047 exist because 044 needs them.

## ADR-044 — Sync Module Revitalization

`packages/core/syncs/` is 1,552 lines that have **never been used**: not exported from `index.js`, zero consumers, zero tests, and `sync.js:74` throws `ReferenceError` on any call (`moduleMap[name]` where `name` is never bound — verified by executing it).

Meanwhile five codebases each built their own sync and none of it is shared:

| Codebase | What it solved |
|---|---|
| `faulkners-nursery--frigg` | structural rendering, relationship resolution by hand |
| `fastspring--integration-service` | relationships — and four ways to get them wrong |
| `freshbooks--frigg` | tombstones, resumable processes, error taxonomy |
| `quo--frigg` | trigger modes, watermarks, fan-out, cancellation |
| `crossbeam--integration-service` | **the hash join, never carried back** |

The finding that reframes the ADR: core's `initialSync()` matches by nested scan — four `filter/some` passes plus a fifth `.find()` *inside* the loop, with an awaited DB write per record. That's ~5·n·m. `GeneralUtil.diff()` in the same repo as crossbeam's own `LHSyncManager` does it in O(n+m), is instrumented with `performance.mark`, and runs at 750,000-record batch scale — and the sync manager next door never calls it.

So this is not export-plus-tests. The reconciliation loop is quadratic and the execution model is unbounded in memory; the three codebases that ran sync in production at volume each proved it by not using it.

Also decided here: storage is a **negotiated tier**, not an assumption. Where the destination upserts on a key we control and the source has delta or events, sync needs no database at all — and the canonical model, rendering split and relationship resolution still earn their keep. Sync objects are valuable without sync records.

## ADR-045 — Capability Traits

044 §2 defines that tier ladder and then can't implement it: an API module `Definition` carries `API`, `getName`, `moduleName`, `modelName`, `requiredAuthMethods`, `env` and stops. Nothing says whether listing supports `modifiedSince` or whether writes are idempotent.

Adds a closed, versioned trait vocabulary to ADR-020 capability declarations — additive, `surface`/`spec`/`implementedBy`/`dependsOn` unchanged. Two authoring paths: inline, or an **OpenAPI Overlay** (stable 1.0.0) so a vendor's published spec is annotated with `x-frigg-traits` rather than forked.

Two properties held firm: `unknown` is the default and means *not declared*, never *not supported*, so a half-annotated library degrades safely; and every trait ships a **conformance probe**, because a module claiming `write.upsertByExternalKey` that can't will silently duplicate every record.

## ADR-046 — Canonical Models

Forkable per-category packages with **declared skew**. The worked case: Aspire's deposit and final draw are *positions in a billing schedule*, not invoice types. A unified accounting schema has nowhere to put that, so it gets rounded off and the invoice is short by a deposit with nobody noticing.

The honest counter-argument is in Negative Consequences and is not fully answered: canonical schema maintenance is where unified API vendors spend most of their engineering, funded by subscription revenue, and a half-maintained canonical model is worse than none.

## ADR-047 — Orchestrator/Worker Queue Pattern

Ratifies the design from 2026-05-14 rather than proposing a new one. Self-publishing workers trip AWS Lambda recursion detection at 16 invocations; **51 Zoho ↔ Quo integrations stalled at 750 contacts**, and one integration had been stuck since 13 April.

`reduce(machine, state, event) → (state', actions[])`, atomic state+outbox, group-FIFO ordering by instance ID, three durability tiers, XState for Tier 3 only. The orchestrator never publishes to the queue it consumes, so the detector stays enabled and never fires — structural, not an account-level override, which preserves detection of genuine runaway loops.

Two consequences called out rather than discovered later: group-FIFO ordering makes a poison message an **instance-wide outage** until the retry cap routes it to DLQ; and the transactional outbox is the easiest thing here to approximate badly.

## Numbering

Claimed **044–047**, above the reconciliation in #644 (028–041), the merged 031/042, and 043 in #646. My first draft used 028–031 against a stale local `next` and collided with four open PRs — caught before pushing.

## Review

Adversarially reviewed by subagents before opening, instructed to falsify rather than confirm. Fixed as a result:

- crossbeam's `diff()` was quoted with `aMap.delete(...)` — a syntax error — and three undeclared variables. Now faithful to source.
- `useDatabase: true` on the extension bundle contradicted ADR-018's explicit rule that a webhook receiver shouldn't pay for a DB connection.
- 047 assigned the orchestrator to the Core Extension contract (017) while using the Integration Extension bundle shape (018). Those are incompatible; extending the Core Extension contract is now named as a prerequisite.
- The quoted `static Config` dropped `name`; the module `Definition` list dropped `moduleName`.
- I claimed ADR-020's surface vocabulary was self-contradictory. It isn't — its Open questions §3 gives the normative enum containing both values. Corrected.
- "Last-writer-wins on duplicate keys" was backwards: crossbeam iterates in reverse, so the *first* array element wins.
- 046 said 23 `v1-ready` API modules; it's 30.
- 044 §9 and 047 §§4/6 specified two different state-machine enforcement points. 044 now defers to 047.

## Related

- #647 — the integration versioning contract these four keep hitting
- #620 — the Integration Versioning draft, which #647 now points at
- #644 — the register reconciliation this numbering sits above
- #646 — ADR lifecycle; supersedes the README convention fixes I'd otherwise have opened separately

https://claude.ai/code/session_01AD57K2eA29SxojjgA5prG1


seanspeaks:
## Redundancy check against the ADRs in #644

Read the pending ADRs the reconciliation surfaced. **No redundancy** — none of 044–047 duplicates an existing or pending ADR. But four cross-references were missing, and one framing was wrong. Pushed in 476e3cc.

### ADR-047 was arguing a settled point

I wrote provider-agnosticism as *"a standing Frigg principle rather than a preference."* Understated — it is **ratified**:

- **ADR-028 Multi-Provider Support** — `Accepted`, 2026-03-02. Commits Frigg to targets beyond AWS.
- **ADR-029 Decouple AWS SDK Dependencies from core** — `Accepted`, 2026-03-03. Defines **`QueueClientInterface`**, which is precisely the dispatch port 047 needs and should ride rather than reinvent.

That materially strengthens 047's rejection of Step Functions: not a taste call, a ratified constraint.

**It also exposes a dependency I had missed.** ADR-029 is Accepted but **unimplemented** — `packages/core/core/Worker.js` and `packages/core/queues/queuer-util.js` still `require('@aws-sdk/client-sqs')` directly on `next` today. So the orchestrator's portability is *contingent on 029 landing*; built before it, the reducer would be portable while its dispatch primitive was not. Now stated in §5.

### ADR-046 was inventing shapes that already have owners

Its agent-assisted authoring proposed "a skill" and an authoring flow from scratch. Both are governed:

- **ADR-036 Skills** names skill subtypes and their lifecycles → 046's skill is now authored per that taxonomy.
- **ADR-037 Agent Pipeline** is the spec-author → build-engineer → adversarial-reviewer flow with schema-validated handoffs → which is exactly the shape 046 §4 describes.

### ADR-044 vs ADR-032 — adjacent, not redundant

**ADR-032 Integration Deletion Data Cleanup** governs Frigg's *own* records on teardown: orphaned `Entity` and `Credential` rows holding still-valid OAuth tokens. ADR-044's tombstones govern *synced third-party* records. Different layers, easily conflated. 044 now says so — and flags that **sync records are a third category ADR-032's cleanup scope should name**, since nothing currently owns them on integration deletion.

### Versioning now points at the right ADR

044, 046 and 047 pointed at ADR-013 for the versioning question. **ADR-030 Integration Versioning** (#620) is the contract ADR-013 explicitly defers to, and is where that question belongs. All three now reference it.

### Checked and found clean

031 / 042 (OAuth concurrency), 033 (Aurora), 034 (API-key login), 035 (app-init), 038–041 (config/secrets) — no overlap with these four.

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+
+## Decision
+
+Revitalize the sync module as a **capability set**: canonical primitives in core, execution and
+workflow as an Integration Extension, with relationships, conflict policy, and lifecycle as
+declarative configuration rather than per-integration code.
+
+### 1. Split primitives from engine
+
+- **`packages/core/syncs/`** keeps `Sync`, `SyncManager`, and the repositories, and **exports them
```
d-klotz, line 259: What repositories?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+    'sync.records.reconcile': {
+        surface: 'sync',
+        spec: { kind: 'json-schema', ref: './specs/sync-config.schema.json' },
+        implementedBy: { kind: 'extension', ref: 'extensions.sync' },
+        dependsOn: [
+            { module: 'aspire', capability: 'accounting.invoice.list' },
+            { module: 'qbo',    capability: 'accounting.invoice.upsert' },
+        ],
+    },
+},
```
d-klotz, line 283: where would that code live?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+  pay for a DB connection."* Queue workers that touch sync records set it per binding.
+
+The extension declares its capability per [ADR-020](./020-capabilities.md), so the sync surface is
+discoverable rather than implied:
+
+```js
+// @friggframework/extension-sync
+capabilities: {
+    'sync.records.reconcile': {
+        surface: 'sync',
```
d-klotz, line 275: I checked ADR-20, looks like "surface" is a free text? Should this be an enum?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+        spec: { kind: 'json-schema', ref: './specs/sync-config.schema.json' },
```
d-klotz, line 276: what would this spec contain? input and output formats?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+| `name`, `keys`, `matchOn`, `moduleMap`, `reverseModuleMap`, `collections`, `relationships`, `skew` | `class XSync { static Config }` | per canonical object |
+| `direction`, `precedence`, `trigger`, `watermark`, `fanOut`, `sortDesc`, `syncObjects`, `tier`, `onSourceDelete`, `onDownstreamDelete` | the extension binding, `Definition.extensions.sync.config` | per integration, per module pair |
+
+```js
+static Definition = {
+    name: 'aspire-qbo',
+    modules: { aspire: { definition: aspire.Definition }, qbo: { definition: qbo.Definition } },
+    extensions: {
+        sync: {
+            extension: require('@friggframework/extension-sync'),
```
d-klotz, line 301: what would enforce that?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+            handlers: { SYNC_TICK: 'onSyncTick' },
+            config: {
+                syncObjects: [InvoiceSync, CustomerSync],
+                direction:   'push',
+                trigger:     'delta',
+                precedence:  { strategy: 'module', master: 'aspire', fallback: 'fillEmpty' },
+                onSourceDelete: 'archive',
```
d-klotz, line 308: what does that mean?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+`bUniques` is the set core has no concept of today. Deletion detection falls out of the primitive
+for free, and `onSourceDelete: 'tombstone' | 'archive' | 'ignore'` becomes configurable rather than
+absent. FreshBooks' `userManuallyDeleted` addresses the mirror-image case — deleted *downstream* —
+and both are needed.
+
+Batched writes replace the per-record `await createSyncDBObject()` currently inside the match loop.
+
+### 5. Fix the constructor gap
+
+`primaryModule` and `secondaryModule` become explicit, validated constructor parameters.
```
d-klotz, line 421: didn't we want to get rid of this concept of primary vs secondary module?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+```js
+class InvoiceSync extends Sync {
+  static Config = {
+    name: 'InvoiceSync',
+    keys: [
+      'number', 'date', 'dueDate', 'currency', 'memo',
+      'classification',   // regular | deposit | intermediate | final — source-derived
+      'amount', 'contractTotal', 'depositAmount',
+      'lines',            // economic lines: { ref, description, amount }
+    ],
```
d-klotz, line 441: Feels weird having to declare Invoice properties in the InvoiceSync configuration. The keys are about the Invoice and not about the syncing mechanism.

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+relationships: {
+  customer: {
+    syncObject:   'CustomerSync',
+    cardinality:  'one',                       // 'one' | 'many'
+    required:     true,
+    from:         (data) => data.customerRef,  // locate parent from child's canonical data
+    cascade:      'ifMissing',                 // 'never' | 'ifMissing' | 'always'
+    onUnresolved: 'defer',                     // 'defer' | 'skip' | 'fail'
+  },
+}
```
d-klotz, line 505: where would this code live? Can you please add more context on the snippets you present?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+  only, never their field data. A parent's phone number changing leaves children untouched; a
+  genuine re-parent updates them. This works only because `relationships` sits outside `keys`.
+
+### 8. Direction and conflict policy
+
+The engine was built for bidirectional sync with one hardcoded rule. That rule becomes one named
+policy among several, and direction becomes explicit:
+
+```js
+direction: 'push',   // 'push' | 'pull' | 'bidirectional' | 'hub'
```
d-klotz, line 536: what does `hub` do?

docs/architecture-decisions/044-sync-module-revitalization.md
```diff
@@ -0,0 +1,756 @@
...
+
+precedence: {
+  strategy: 'module',                  // 'module' | 'field' | 'recency' | 'custom'
+  master:   'aspire',
```
d-klotz, line 540: how would master behave if we choose direction `bidirectional`?

docs/architecture-decisions/044-sync-module-revitalization.md
d-klotz: ok, it's right down there

docs/architecture-decisions/045-capability-traits.md
```diff
@@ -0,0 +1,242 @@
...
+// @friggframework/api-module-hubspot/definition.js
+const Definition = {
+    API: Api,
+    moduleName: config.name,
+    // …auth-centric keys unchanged…
+    capabilities: {
+        'crm.contact.list': {
+            surface: 'data-read',
+            spec: { kind: 'openapi', ref: './specs/hubspot-openapi.yaml#/paths/~1crm~1v3~1objects~1contacts/get' },
+            implementedBy: { kind: 'api-class-method', ref: 'HubSpotApi.listContacts' },
```
d-klotz, line 65: what is `api-class-method`? `api.js` ?

docs/architecture-decisions/045-capability-traits.md
```diff
@@ -0,0 +1,242 @@
...
+        },
+        'crm.contact.upsert': {
+            surface: 'data-write',
+            spec: { kind: 'openapi', ref: './specs/hubspot-openapi.yaml#/paths/~1crm~1v3~1objects~1contacts~1batch~1upsert/post' },
+            implementedBy: { kind: 'api-class-method', ref: 'HubSpotApi.upsertContacts' },
+            traits: {
+                'write.upsertByExternalKey': 'idProperty',
+                'write.batchMax': 100,
+                'write.partialUpdate': true,
+            },
```
d-klotz, line 82: Isn't that basically duplicating the API spec?

docs/architecture-decisions/045-capability-traits.md
```diff
@@ -0,0 +1,242 @@
...
+        },
```
d-klotz, line 83: Should we version the definition as well? What if these traits change?

docs/architecture-decisions/045-capability-traits.md
```diff
@@ -0,0 +1,242 @@
...
+```yaml
+overlay: 1.0.0
+info: { title: Frigg traits for HubSpot CRM, version: 1.0.0 }
+actions:
+  - target: $.paths['/crm/v3/objects/contacts'].get
+    update:
+      x-frigg-traits:
+        delta.modifiedSince: true
+        read.pageMax: 100
+```
```
d-klotz, line 144: I don't understand why we need this

docs/architecture-decisions/045-capability-traits.md
```diff
@@ -0,0 +1,242 @@
...
+verification requirement. It does **not** define what any consumer does with the answer. The sync
+storage-tier ladder lives in ADR-044 §2. Future consumers bring their own interpretation.
+
+## Consequences
+
+### Positive
+
+- Stateless-first becomes achievable rather than aspirational. The engine can *know* when it may
+  store nothing, instead of defensively persisting a record per pair forever.
+- Assumptions currently buried in integration code become declared, reviewable, and testable.
```
d-klotz, line 186: how can we trust that what's declared is indeed what's implemented?

docs/architecture-decisions/046-canonical-models.md
```diff
@@ -0,0 +1,203 @@
...
+the canonical fields, visible to maps that want them and ignorable by maps that do not.
+
+```js
+static Config = {
+    name:    'InvoiceSync',
+    keys:    ['number', 'date', 'amount', 'classification'],
+    matchOn: ['number'],
+    skew: {
+        aspire: {
+            keys:   ['billingScheduleSortOrder', 'opportunityId'],
```
d-klotz, line 93: how is nesting handled?

docs/architecture-decisions/046-canonical-models.md
```diff
@@ -0,0 +1,203 @@
...
+            hashed: false,   // carried and preserved, but excluded from the change hash
+        },
+    },
+};
```
d-klotz, line 97: where would this code live? In the integration?

PR #651, opened 2026-09-22
feat(core): add queryMappings to IntegrationMappingRepository

d-klotz:
Adds a filtered, paginated query to the integration mapping repository, so an integration can page through its mappings in the database — PostgreSQL, MongoDB or DocumentDB — instead of loading every row with `findMappingsByIntegration`.

## What this enables
- Every Frigg integration already has the same `IntegrationMapping` table. Each row has an `integrationId`, a `sourceId`, and a `mapping` field where the integration stores any JSON it wants.
- **Before:** the only way to read those rows was `findMappingsByIntegration`, which loads *all* of them into the Lambda. At 10,000 rows that took 6.5 s and 2 GB of memory.
- **Now:** any integration can ask for "page 3 of my mappings, where some JSON field exists or has one of these values, sorted by another JSON field, plus the total count". The database does the filtering, sorting and counting, on PostgreSQL, MongoDB and DocumentDB.
- **Example:** a Stripe↔Salesforce integration that stores `{ lastSync: { status, at } }` can show a "failed records" table sorted by time. It passes `mapping.lastSync.status` and `mapping.lastSync.at` as the paths; the method accepts any path made of plain identifiers.

```js
const { mappings, total } = await integrationMappingRepository.queryMappings(integrationId, {
    where: [{ path: 'mapping.lastSync.status', op: 'in', value: ['failed'] }],
    orderBy: { path: 'mapping.lastSync.at', direction: 'desc' },
    skip: 50,
    take: 25,
});
```

## Why
Integrations that show their sync state to users — for example a "recent syncs" or "failed records" list — need a filtered, sorted page of their mappings plus a total. Today the only way is `findMappingsByIntegration`, which loads every mapping row into the Lambda and leaves filtering, sorting and paging to the integration. With 10⁴+ mappings per integration, and mappings that carry audit data such as change logs, that load alone can exceed the API Gateway / Lambda time limit and use gigabytes of memory.

## API
`queryMappings(integrationId, { where = [], orderBy, skip = 0, take, omit = [] })` → `{ mappings, total }`, rows shaped like `findMappingsByIntegration`.
- `where`: AND of conditions, one level of `{ anyOf }`. `exists` / `notExists` / `in` on `mapping.<path>` (`exists` excludes JSON null); `notStartsWith` on `sourceId`.
- `orderBy`: `{ path: 'mapping.<path>', direction }`, nulls last, ties by `id` in the same direction.
- `take` 1–500, `skip` ≥ 0, `omit`: top-level mapping keys to leave out.

## Safety
- Input is validated before any DB call; identifier-only path segments.
- Postgres: fixed SQL text via the existing `bind()` + `$queryRawUnsafe` pattern. Every value, JSON paths included, is a bound parameter; sort direction comes from a whitelist.
- Only object mappings match, so legacy ciphertext rows never match.
- Encryption: throws when `IntegrationMapping.mapping` is encrypted on write (the query filters JSON in the database). With encryption disabled (`STAGE=dev/test/local`) or the field opted out via `appDefinition.encryption.disable`, it runs. It never decrypts or returns ciphertext.
- MongoDB / DocumentDB: an aggregation built by one shared stage builder (`integration-mapping-query-pipeline.js`). Filters run in `$expr`, so paths only resolve through objects and `in` only matches strings, like Postgres (plain query operators would reach into arrays). Caller values go through `$literal`; `notStartsWith` uses `$substrCP`, no regex.
- The legacy `IntegrationMappingRepository` class exported from `index.js` still inherits the port's "not supported" error: delegating would change its row shape (integer ids on Postgres, unparseable string `_id`s on DocumentDB).
- No schema change, no new index: `@@index([integrationId])` narrows the scan.

## Test plan
- [x] `integration-mapping-query.test.js` (89) and `integration-mapping-repository-query.test.js` (29) pass
- [x] eslint: 0 errors on changed files
- [x] Real Postgres (throwaway container, real migration's table): every page and total matched an in-memory reference filter across directions, status filters, sort directions and skip/take pairs; ciphertext / SQL NULL / JSON array rows never match; quote and brace characters stay data; `EXPLAIN` uses the `(integrationId, sourceId)` index
- [ ] Core suite: no new failures (existing failures need live databases, AES keys or a missing module)

<details>
  <summary>📦 Published PR as canary version: <code>2.0.0--canary.651.a7237ae.0</code></summary>
  <br />

  ```bash
  npm install @friggframework/admin-scripts@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/core@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/devtools@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/eslint-config@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/prettier-config@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/schemas@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/serverless-plugin@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/test@2.0.0--canary.651.a7237ae.0
  npm install @friggframework/ui@2.0.0--canary.651.a7237ae.0
  # or 
  yarn add @friggframework/admin-scripts@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/core@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/devtools@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/eslint-config@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/prettier-config@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/schemas@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/serverless-plugin@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/test@2.0.0--canary.651.a7237ae.0
  yarn add @friggframework/ui@2.0.0--canary.651.a7237ae.0
  ```
</details>

## Performance
Loading every mapping row and filtering in memory: 10⁴ rows took 6.5 s and added 2.16 GB RSS. `queryMappings` returns one page (about 130 kB for 200 rows), filtered and ordered in Postgres, in **one statement**: a `matched` CTE, `(SELECT COUNT(*) FROM matched)` for the total, and the page `LEFT JOIN`ed onto one row, so an empty page still returns the total. Postgres 16, ~4 kB JSON per row, median of 5:

| rows | case | old (`COUNT(*) OVER ()` + fallback) | new (CTE) |
|---|---|---|---|
| 10⁴ | first page | 390 ms | 346 ms |
| 10⁴ | past the end | 656 ms | 330 ms |
| 10⁵ | first page | 3148 ms | 2625 ms |
| 10⁵ | past the end | 5704 ms | 2601 ms |

Each call is still O(rows of the integration × mapping size): JSON paths are evaluated per row and there is no JSON index — fine at ~10⁴ rows per integration, slower at 10⁵+. The cost model is in `packages/core/CLAUDE.md`.

## Review follow-up
An independent max-effort review approved with minor follow-ups, all addressed:
- The encryption guard also refuses when a nested `mapping.*` path is encrypted (the error names the paths to opt out).
- Page and total in one statement (`COUNT(*) OVER ()`), with a count-only fallback just for an empty page after the first.
- Input bounds: at most 500 `in` values and 20 `where` conditions (anyOf members included), validated before any SQL.
- The legacy `IntegrationMappingRepository` exported from `index.js` refuses `queryMappings` clearly.
- A passed encryption check is kept per repository instance, so a prod stage without keys logs its WARN once, not per call.
- Cost model stated in the interface JSDoc.

Test plan update: query tests 136/136; the core suite's failing set is identical to the PR's original head (missing `map-integration-dto` path, live Mongo/DocumentDB, AES/KMS keys); real Postgres 16 semantics re-checked, pages and totals unchanged. CI on `next` itself fails the same way (MongoMemoryServer `libcrypto.so.1.1`, devtools suites), so CI gives no signal for core here.

## Code review follow-up (`01c2cd80`..`9aaeff61`)
Standards + spec review, all addressed:
- **Encryption decision out of the repository:** `getMappingFieldsEncryptedOnWrite()` in `database/encryption/integration-mapping-encryption.js`. A pass is kept per process (supersedes "per repository instance" above), a refusal re-checks every call; `resetMappingEncryptionCheck()` for tests.
- **One statement per call** (see Performance) — no second count statement, no mutation of returned rows.
- **Adapter parity:** Mongo, DocumentDB and legacy no longer override `queryMappings` just to throw; they inherit the port's error, like `countByIntegrationIds` (ADR-010). *(Superseded: Mongo and DocumentDB now implement it — see below.)*
- **One entry per operator on each side:** the validator emits normalised `{ field, path?, op, value? }` and returns `asc`/`desc`; the adapter maps ops and SQL keywords from its own tables. Unused `SEGMENT_REGEX` export dropped.
- **Docs:** `queryMappings` in `packages/core/CLAUDE.md` (with the cost model) and the opt-out requirement in `database/encryption/README.md`; JSDoc reduced to the contract.
- **Harness:** `frigg-local-test` ran against a throwaway Postgres (port 55432): `integration-lifecycle` 4/4, plus a `queryMappings` scenario through a hydrated `IntegrationBase` subclass 7/7.

Tests: affected suites 230 pass; the 4 failures (encryption-schema-registry Credential fields, DocumentDB encryption) fail identically on `next`. Lint: 0 errors. 576 query shapes from a real downstream integration validate and build SQL. Test fixtures use neutral values (`bc0e60a3`).

## All supported databases (`a11dab3b`..`8ef5a7a4`)
`queryMappings` now has the same contract on PostgreSQL, MongoDB and DocumentDB:
- **MongoDB:** one aggregate; `$facet` returns the page and `$count` together. Matches `integrationId` as an ObjectId; rows shaped like `findMappingsByIntegration`.
- **DocumentDB:** `$facet` is not supported on 4.0/5.0, so the page and a `$count` run as two aggregates in parallel (the total is not from the same snapshot as the page). The page ends with `$sort` again, because DocumentDB only keeps sort order when `$sort` is last. Uses only stages/operators AWS lists as supported on 4.0 and 5.0 (no `$getField`, `$set`, `$unset`); `allowDiskUse` for large integrations.
- **Sort:** nulls last in both directions, then JSON type order as in Postgres, then value, then `_id` in the same direction.
- **Encryption:** the refusal moved into the encryption module and is shared by all three adapters.
- **Bug fix needed for DocumentDB:** `DocumentDBEncryptionService.encryptFields` ignored `appDefinition.encryption.disable`, so an opted-out field was still written encrypted (the Prisma extension already honoured it). Writes now use `getFieldsToEncryptOnWrite`; reads still decrypt opted-out fields, so older rows stay readable. **This changes DocumentDB writes for apps that use `encryption.disable`.**
- **Indexes:** the Mongo schema's `@@index([integrationId])` narrows the scan (checked with `explain`); no schema change.

Known differences: Postgres sorts strings by database collation, Mongo/DocumentDB by code point; arrays/objects at the sort path fall back to `_id` order on Mongo; a MongoDB page (after `omit`) must fit in one 16 MB document.

Tests: 17 MongoDB and 13 DocumentDB unit tests pin the exact pipelines; a parity suite runs 461 cases × 3 adapters (MongoDB, the DocumentDB pipeline on MongoDB, PostgreSQL) against real MongoDB 7 and PostgreSQL 16 — all pass. It is skipped unless `QUERY_MAPPINGS_PARITY_MONGO_URL` / `QUERY_MAPPINGS_PARITY_POSTGRES_URL` are set. Not run on a real DocumentDB cluster. Mapping, encryption and DocumentDB suites: 176 pass; eslint 0 errors. MongoDB 7, ~4 kB mappings: ~0.05–0.1 s per call at 10⁴ rows per integration, ~0.5–0.7 s at 10⁵.


packages/core/integrations/repositories/integration-mapping-repository-postgres.js
```diff
@@ -216,6 +238,101 @@ class IntegrationMappingRepositoryPostgres extends IntegrationMappingRepositoryI
...
+            SELECT "id", "integrationId", "sourceId", ${mappingSql}, "createdAt", "updatedAt",
+                (SELECT COUNT(*)::int FROM "matched") AS "__total"
+            FROM (VALUES (1)) AS "one"
+            LEFT JOIN "page" ON true
+            ORDER BY ${orderSql}
+        `;
+        const rows = await this.prisma.$queryRawUnsafe(sql, ...params);
+
+        return {
+            mappings: rows
```
roboli: **Legacy nested-path ciphertext is returned as-is (reproduced).**

The PR says `queryMappings` "never decrypts or returns ciphertext". That holds when the whole `mapping` was encrypted (those rows are strings, so `jsonb_typeof = 'object'` excludes them), but not when a custom `encryption.schema` encrypted a *nested* path, e.g. `mapping.apiSecret`, before the app opted out. Those older rows are still objects, so they match, and because this adapter reads through `$queryRawUnsafe` (bypassing the Prisma encryption extension) the ciphertext comes back untouched.

Reproduced against real PostgreSQL 16 and MongoDB 7 (AES encryption on):

1. Custom schema `IntegrationMapping: ['mapping.apiSecret']`, opt-out `['mapping']`. Save `{ status: 'failed', apiSecret: 'my-real-secret' }` → stored with `status` plain, `apiSecret` encrypted.
2. New process, opt-out `['mapping', 'mapping.apiSecret']`. Read the row both ways:

```
            findMappingsByIntegration   queryMappings
MongoDB     "my-real-secret"            bG9jYWwtdGVzdC1rZXk=:74a… (ciphertext)
DocumentDB  "my-real-secret"            "my-real-secret"
PostgreSQL  "my-real-secret"            bG9jYWwtdGVzdC1rZXk=:89f… (ciphertext)
```

So the three adapters disagree on the same row, and Postgres/Mongo disagree with `findMappingsByIntegration`. Reach is narrow (default config encrypts the whole `mapping`, which is handled correctly), but it's a real gap.

Suggestion: run the returned rows through the same decrypt-on-read step DocumentDB already uses here (FieldEncryptionService / `getFieldsToDecryptOnRead`), and add this scenario as a test. If you'd rather not, a note in the encryption README plus correcting the PR description would do.

packages/core/integrations/repositories/integration-mapping-repository-mongo.js
```diff
@@ -155,6 +173,57 @@ class IntegrationMappingRepositoryMongo extends IntegrationMappingRepositoryInte
...
+                    },
+                },
+            ],
+            cursor: {},
+            allowDiskUse: true,
+        });
+        const [{ mappings, total }] = result.cursor.firstBatch;
+
+        return {
+            mappings: mappings.map((doc) => this._fromRawMapping(doc)),
```
roboli: Same issue as the PostgreSQL comment: `$runCommandRaw` bypasses the Prisma encryption extension, so a legacy row whose nested `mapping.*` path was encrypted before the opt-out comes back with the ciphertext (reproduced on MongoDB 7). The DocumentDB adapter decrypts here and returns the plain value.

packages/core/database/documentdb-encryption-service.js
```diff
@@ -107,7 +111,7 @@ class DocumentDBEncryptionService {
         }
 
         // Get encrypted fields from registry
-        const encryptedFieldsConfig = getEncryptedFields(modelName);
+        const encryptedFieldsConfig = getFieldsToEncryptOnWrite(modelName);
```
roboli: Question / suggestion on this fix: it's the right change, but it changes what existing DocumentDB apps write. Any app on DocumentDB that already has `encryption.disable` will start storing those fields as plain JSON on its next deploy, where before this PR they were still encrypted.

As far as I can tell, none of the DocumentDB code (this change or the new `queryMappings` pipeline) has run against a real DocumentDB cluster: the unit tests mock `$runCommandRaw`, and the parity suite's DocumentDB leg runs on MongoDB. Would it be possible to:

- run the parity suite once against a dev DocumentDB cluster (`QUERY_MAPPINGS_PARITY_MONGO_URL` via the bastion, TLS + `retryWrites=false`), and
- check with the canary on a dev DocumentDB app using `encryption.disable` that new writes are plain and older encrypted rows still read back?

And either way, could the write-side change get an explicit callout in the release notes?

roboli (approved):
Nice work. The validation, bound parameters and shared operator tables make this easy to trust, and I checked on real MongoDB 7 that Mongo `queryMappings` returns numbers (int32/int64/double, even Decimal128/Date) identically to `findMappingsByIntegration`.

Two things inline: a reproduced edge case where legacy nested-path ciphertext is returned undecrypted on Postgres/Mongo, and a question about validating the DocumentDB write change on a real cluster. Neither blocks from my side, approving.

packages/core/integrations/repositories/integration-mapping-repository-postgres.js
d-klotz: Good catch, fixed in 4c3f1fc7. Postgres and Mongo rows now go through the same decrypt-on-read as the Prisma extension (`decryptQueriedMappings`). It's a no-op when encryption is off. Your scenario is in the unit tests and the parity suite, and all 3 adapters now return the plain value, same as `findMappingsByIntegration`. Filters and ordering on that path still see ciphertext until the row is rewritten; that's in the README.

packages/core/integrations/repositories/integration-mapping-repository-mongo.js
d-klotz: Same fix as the Postgres thread (4c3f1fc7).

packages/core/database/documentdb-encryption-service.js
d-klotz: Agreed, added a release-notes callout to the description. Haven't run it on a real DocumentDB cluster yet; will run the parity suite through the bastion and check the canary on a dev DocumentDB app with `encryption.disable` before merge.

packages/core/integrations/repositories/integration-mapping-repository-interface.js
```diff
@@ -52,6 +52,43 @@ class IntegrationMappingRepositoryInterface {
...
+     * @param {number} [query.skip=0] - Rows to skip (integer ≥ 0)
+     * @param {number} query.take - Page size (integer 1–500)
+     * @param {string[]} [query.omit=[]] - Top-level mapping keys to leave out of
+     *   the returned rows; such projected rows must not be written back
+     * @returns {Promise<{mappings: Array<Object>, total: number}>} The page, and
+     *   the number of rows matching `where`
+     * @abstract
+     */
+    async queryMappings(integrationId, query) {
+        throw new Error('Method queryMappings must be implemented by subclass');
```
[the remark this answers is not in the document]
d-klotz: Intentional. The legacy class keeps the refusal, because delegating would change its row shape (integer ids on Postgres, string `_id`s that DocumentDB's Prisma model can't parse). `IntegrationBase` uses the factory, which returns the adapter with `queryMappings`.

PR #653, opened 2026-09-23
feat(core): tell queue handlers their last delivery attempt

d-klotz:
Queue handlers now learn whether an SQS delivery is their last attempt, so an integration can end a run or count lost work on the final try instead of letting the message vanish into the DLQ.

## What problem this fixes
- **Temporary errors are retried by SQS.** When a queue handler hits a temporary error (a 429 after the requester's backoff, a 5xx, a network error), it rethrows and SQS delivers the message again 30 minutes later. The integration queue allows **3 deliveries**. After the 3rd failure SQS moves the message to the DLQ, and the DLQ processor only logs it.
- **The handler can't tell which delivery it is on.** Core reads the receive count (it even logs it) but never passes it to the handler. So the handler treats its last chance exactly like its first: it rethrows, and the work is silently lost.
- **For a sync split into many messages, one lost message stalls the whole run.** The run is never finished and never failed: it stays "running" until something treats it as stale, and nobody is told.
- **Example:** a sync of 5,000 records runs as 50 batch messages, and the API it calls rate-limits for two hours. Batch 17 fails on all 3 deliveries and lands in the DLQ. The other 49 batches finish, and the run waits forever for batch 17.
- **With this PR** the handler gets `delivery.isLastAttempt`. On the last try it can do something useful instead of rethrowing into the void: mark the run failed (the user sees it and can retry), or count the batch's records as failed so the run completes.
- **Nothing else changes.** Core still rethrows and discards the same errors, and SQS and the DLQ behave exactly as before. This PR only passes on information core already has.

## API
Queue handlers get `{ data, context, delivery }`. Handlers that only read `{ data, context }` are unchanged.
- `delivery.receiveCount` — `number | undefined`, from `ApproximateReceiveCount`
- `delivery.maxReceiveCount` — `number | undefined`, from `FRIGG_QUEUE_MAX_RECEIVE_COUNT`
- `delivery.isLastAttempt` — `true` only when both are known and `receiveCount >= maxReceiveCount`

Present for every SQS-dispatched handler (sync events, queued webhooks, scheduled jobs); absent for HTTP routes and `this.on` events.

```js
async processBatch({ data, delivery }) {
    try { await this.syncPage(data); }
    catch (error) {
        if (delivery?.isLastAttempt) await this.failRun(data.processId, error);
        throw error;
    }
}
```

## Changes
- `852a45d6` — `core/queues/queue-delivery.js` (`readQueueDelivery`, `INTEGRATION_QUEUE_MAX_RECEIVE_COUNT = 3`); `Worker.js` → queue worker → `integration-event-dispatcher.js` pass `delivery` through.
- `02d9d897` — the integration builder takes the redrive `maxReceiveCount` from the core constant and sets `FRIGG_QUEUE_MAX_RECEIVE_COUNT` on `<name>QueueWorker`. One source of truth; the runtime has no copy of the number.
- `af34bd7c` — docs: `packages/core/CLAUDE.md`, `core/core/CLAUDE.md`, `docs/guides/INTEGRATION-PATTERNS.md`.

No behaviour change otherwise: core still rethrows, still discards halt errors (4xx except 408/429), never deletes, re-sends or changes visibility; the DLQ processor is untouched.

## Test plan
- [x] core: 8 suites, 94 pass — receive counts 1/2/3 → `isLastAttempt` false/false/true, missing count, unknown max, a handler reading only `{ data, context }`
- [x] devtools: integration + env suites, 102 pass — redrive value and env var come from the core constant; an external queue gets no env var
- [x] eslint: 0 errors
- [ ] `infrastructure-composer.test.js`: the same 52 tests fail with and without this change

## Caveats
- `isLastAttempt: true` is reliable; its absence is not proof the message survives. A Lambda timeout or crash on the last attempt skips the handler's `catch`, and throttled receives can dead-letter a message the handler never saw as last.
- `ownership.queue: 'external'` queues get no `FRIGG_QUEUE_MAX_RECEIVE_COUNT` (the stack doesn't know their redrive policy), so `isLastAttempt` stays false there.
- Admin-script queues still hardcode `maxReceiveCount: 3`; left alone.
- Needs a redeploy of the stack for the env var to reach the queue worker.

<details>
  <summary>📦 Published PR as canary version: <code>2.0.0--canary.653.8213092.0</code></summary>
  <br />

  ```bash
  npm install @friggframework/admin-scripts@2.0.0--canary.653.8213092.0
  npm install @friggframework/core@2.0.0--canary.653.8213092.0
  npm install @friggframework/devtools@2.0.0--canary.653.8213092.0
  npm install @friggframework/eslint-config@2.0.0--canary.653.8213092.0
  npm install @friggframework/prettier-config@2.0.0--canary.653.8213092.0
  npm install @friggframework/schemas@2.0.0--canary.653.8213092.0
  npm install @friggframework/serverless-plugin@2.0.0--canary.653.8213092.0
  npm install @friggframework/test@2.0.0--canary.653.8213092.0
  npm install @friggframework/ui@2.0.0--canary.653.8213092.0
  # or 
  yarn add @friggframework/admin-scripts@2.0.0--canary.653.8213092.0
  yarn add @friggframework/core@2.0.0--canary.653.8213092.0
  yarn add @friggframework/devtools@2.0.0--canary.653.8213092.0
  yarn add @friggframework/eslint-config@2.0.0--canary.653.8213092.0
  yarn add @friggframework/prettier-config@2.0.0--canary.653.8213092.0
  yarn add @friggframework/schemas@2.0.0--canary.653.8213092.0
  yarn add @friggframework/serverless-plugin@2.0.0--canary.653.8213092.0
  yarn add @friggframework/test@2.0.0--canary.653.8213092.0
  yarn add @friggframework/ui@2.0.0--canary.653.8213092.0
  ```
</details>


packages/devtools/infrastructure/domains/integration/integration-builder.js
```diff
@@ -32,6 +32,10 @@ const {
 const {
     nestedNodeModulesExcludes,
 } = require('../shared/utilities/nested-node-modules');
+const {
+    INTEGRATION_QUEUE_MAX_RECEIVE_COUNT,
+    QUEUE_MAX_RECEIVE_COUNT_ENV,
+} = require('@friggframework/core/queues/queue-delivery');
```
roboli: Suggestion for the release notes: this makes devtools depend on a core file that only exists from this version on. An app that upgrades `@friggframework/devtools` but keeps an older `@friggframework/core` will fail at `frigg deploy` with `Cannot find module '@friggframework/core/queues/queue-delivery'`. npm won't catch the mismatch, since devtools declares `"@friggframework/core": "^2.0.0-next.0"`, which older cores still satisfy.

It fails loudly at deploy time and the fix is just upgrading core, so I don't think it needs a code change. A line in the release notes saying core and devtools should be upgraded together would cover it.

packages/core/handlers/backend-utils.js
```diff
@@ -254,6 +254,7 @@ const createQueueWorker = (integrationClass) => {
                     event: params.event,
                     data: params.data,
                     context: context,
+                    delivery,
```
roboli: Small addition for the Caveats section: `delivery` only reaches the handler once hydration succeeds. If loading the integration throws before dispatch, e.g. `Process not found` from `loadIntegrationForProcess`, or credentials that can't be loaded/decrypted, it fails the same way on every attempt, the handler never sees `isLastAttempt`, and the message still dead-letters silently, so the run can still hang in that case.

The handler often couldn't act there anyway (no process to fail, or no DB to write to), and fixing it properly would mean core marking the run failed itself, which is out of scope here. Just worth documenting next to the Lambda-timeout caveat.

roboli (approved):
Clean, well-scoped change: it passes on information core already had, keeps a single source of truth for the receive count, and doesn't change SQS/DLQ behaviour. I checked the builder ordering (the queue worker is defined before the env var is added), `batchSize: 1` (no neighbours burning retries), and that sync events, webhooks and scheduled jobs all reach `dispatchJob` with `delivery`.

Two small, non-blocking notes inline: a release-note suggestion for the new devtools → core import, and one more caveat worth documenting. Approving.

packages/devtools/infrastructure/domains/integration/integration-builder.js
d-klotz: Makes sense, added to the release notes in the description.

packages/core/handlers/backend-utils.js
d-klotz: Good point, added next to the Lambda-timeout caveat.

